PR #7, opened 2022-10-19
SYS-2747 Use central pot to pay out rewards

nahuseyoum:
This PR builds on top of [PR#6](https://github.com/Aventus-Network-Services/avn-parachain/pull/6) and uses the new reward pot to payout rewards. Tests have also been updated.


pallets/parachain-staking/src/lib.rs
```diff
@@ -577,6 +584,16 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn locked_era_payout)]
+	/// Storage value that holds the total amount of payouts we are waiting to take out of this pallet's pot.
+	pub type LockedEraPayout<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
```
nahuseyoum, line 590: Kept the name because we will rename `round` to `era` for consistency

fluorostani (approved):

pallets/parachain-staking/src/lib.rs
```diff
@@ -1315,19 +1332,30 @@ pub mod pallet {
...
-			}
+
+		/// Compute total reward for round based on the amount in the reward pot
+		fn compute_total_reward_to_pay() -> BalanceOf<T> {
+			let total_unpaid_reward_amount = Self::reward_pot();
+			let mut payout = total_unpaid_reward_amount.checked_sub(&Self::locked_era_payout()).or_else(|| {
+				log::error!("💔 Error calculating era payout. Not enough funds in total_unpaid_reward_amount.");
+
+				//This is a bit strange but since we are dealing with money, log it.
+				Self::deposit_event(Event::NotEnoughFundsForEraPayment {reward_pot_balance: total_unpaid_reward_amount});
```
alexpArtos, line 1343: Is this error used for anything else? If not, then it feels like we should also report the locked payout to help us find why that would be higher than the unpaid reward.

alexpArtos (approved):

pallets/parachain-staking/src/lib.rs
nahuseyoum: We could, but if you see this event then you can query the balance of `locked_era_payout` at the time the event was raised. Having the `reward_pot_balance` and `locked_era_payout` on the event will not be enough to find the reason.

thadouk (commented):
approved

PR #10, opened 2022-10-20
SYS-2779 remove migrations

nahuseyoum:
This PR removes the old migration code


fluorostani (approved):

RuiSMagalhaes (approved):

thadouk (approved):
If we had the removal before the  rename PR the rename PR would be slightly smaller :)

PR #11, opened 2022-10-20
SYS-2779 rename `round` to `era`

nahuseyoum:
This PR is a pure rename of `round` to `era` with no additional functionality. This rename is useful because it matches the terminology used by substrate's staking pallet and our offchain products.


fluorostani (approved):

pallets/parachain-staking/src/nomination_requests.rs
```diff
@@ -47,18 +47,18 @@ impl<Balance: Copy> NominationAction<Balance> {
 }
 
 /// Represents a scheduled request that define a [NominationAction]. The request is executable
-/// iff the provided [RoundIndex] is achieved.
+/// iff the provided [EraIndex] is achieved.
```
thadouk, line 50: existing typo?
```suggestion
/// if the provided [EraIndex] is achieved.
```

thadouk (approved):
Does this Era match the Substrate Era in terms of time?
Meaning is 1 Era in the staking pallet starts and end in an era of the framework & the sessions?
If not this could create some confusion.

nahuseyoum:
> Does this Era match the Substrate Era in terms of time? Meaning is 1 Era in the staking pallet starts and end in an era of the framework & the sessions? If not this could create some confusion.

We don't use substrate's staking pallet on our parachain so there shouldn't be any overlap there. In terms of defining how long an era is, its configurable.

pallets/parachain-staking/src/nomination_requests.rs
nahuseyoum: Its not a typo, it means "if and only if"

alexpArtos: It's a standard in maths books

alexpArtos: But I do agree that it is ambiguous for anyone who does not have that background, and that we should be explicit and avoid abbreviations where possible

alexpArtos (approved):

alexpArtos (approved):

thadouk (approved):

fluorostani (approved):

PR #15, opened 2022-10-25
SYS-2752 format code

nahuseyoum:
This PR adds a `rustfmt` file and runs it to format the code. It also adds a `gitignore` and root `cargo.toml` file.


pallets/parachain-staking/Cargo.lock
```diff
@@ -0,0 +1,3529 @@
+# This file is automatically @generated by Cargo.
```
thadouk, line 1: I don't think we should check in the Cargo.lock file on a module.

.gitignore
```diff
@@ -0,0 +1,20 @@
...
+**/*.rs.bk
+
+# Local chain databases (defalut location)
+**/chains/
+
+# The cache for chain data in container
+.local
+
+# The cache for docker container dependency
+/.cargo/config
```
thadouk, line 14: what container is that? I don't see any dockerfiles in the repo or the PR
If this is part of another PR can we add that there please?

thadouk (approved):

RuiSMagalhaes (dismissed):

PR #35, opened 2022-10-31
SYS-2757 process collator payouts

nahuseyoum:
Pre-requisite: 
- [ ] Open source AvN pallet

This PR adds the logic of paying out collators for producing blocks. 
If the `growth period` received from T1 matches, the reward is paid for collators that produced blocks for that period, according to their performance. If the growth period does not match, current collators will get paid equally.

Next PR will remove storage data for processed payouts to reduce state bloat.


pallets/parachain-staking/src/lib.rs
```diff
@@ -1555,15 +1583,82 @@ pub mod pallet {
...
+            return current_scores
+                .into_iter()
+                .map(|(acc, pts)| CollatorScore::new(acc, pts))
+                .collect()
+        }
+
+        pub fn payout_collators(amount: BalanceOf<T>, growth_period: u32) -> DispatchResult {
+            // We do not validate anything here, we trust that the instruction to pay from T1 is ok
+
+            let pay = |collator_address: T::AccountId, amount: BalanceOf<T>| {
```
alexpArtos, line 1628: Why do we need to define this function at this level, and not outside? 
I think it makes it harder to understand what is going on, and don't understand its justification.

pallets/parachain-staking/src/lib.rs
```diff
@@ -1555,15 +1583,82 @@ pub mod pallet {
...
+                    Self::deposit_event(Event::CollatorPaid {
+                        account: collator_address,
+                        amount: amount_paid.peek(),
+                        period: growth_period,
+                    });
+                }
+            };
+
+            if <Growth<T>>::contains_key(growth_period) {
+                // get the list of candidates that earned points from `growth_period`
```
alexpArtos, line 1641: We should prefer to have well-named functions instead of code comments

pallets/parachain-staking/src/lib.rs
```diff
@@ -1555,15 +1583,82 @@ pub mod pallet {
...
+                let growth_data = <Growth<T>>::get(growth_period);
+                for collator_data in growth_data.collator_scores {
+                    let percent =
+                        Perbill::from_rational(collator_data.points, growth_data.total_points);
+                    pay(collator_data.collator, percent * amount);
+
+                    // TODO: remove processed records to reduce state bloat
+                }
+            } else {
+                // get the list of current candidates because we there is no way of knowing who they were
```
alexpArtos, line 1651:
```suggestion
                // get the list of current candidates because there is no way of knowing who they were
```

alexpArtos (approved):

pallets/parachain-staking/src/lib.rs
nahuseyoum: Its a single scope function so there is no need for it to be defined outside. Using a closure allows us to encapsulate the logic in one place.

thadouk (approved):

fluorostani (approved):

pallets/parachain-staking/src/lib.rs
```diff
@@ -1555,15 +1583,82 @@ pub mod pallet {
...
+
+            if <Growth<T>>::contains_key(growth_period) {
+                // get the list of candidates that earned points from `growth_period`
+                let growth_data = <Growth<T>>::get(growth_period);
+                for collator_data in growth_data.collator_scores {
+                    let percent =
+                        Perbill::from_rational(collator_data.points, growth_data.total_points);
+                    pay(collator_data.collator, percent * amount);
+
+                    // TODO: remove processed records to reduce state bloat
```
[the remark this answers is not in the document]
nahuseyoum: Yes, the update is coming in a new PR

PR #36, opened 2022-10-31
SYS-1642, SYS-2757 add primitives

nahuseyoum:
This PR adds avn primitives.


primitives/avn-common/src/event_types.rs
```diff
@@ -0,0 +1,757 @@
...
+use serde::{Deserialize, Serialize};
+use sp_core::{H160, H256, H512, U256};
+use sp_runtime::{scale_info::TypeInfo, traits::Member, DispatchResult};
+use sp_std::{convert::TryInto, vec::Vec};
+
+// ================================= Events Types ====================================
+
+const WORD_LENGTH: usize = 32; // basic word type for Ethereum is 32 bytes
+const HALF_WORD_LENGTH: usize = 16; // needed for creating a u128
+const TWENTY_FOUR_BYTES: usize = 24; // needed for creating a u64
```
alexpArtos, line 13: This comment seems odd. A u64 is 8 bytes. At any rate, it's always smaller than a u128. Why then would it need 24 bytes?

primitives/avn-common/src/event_types.rs
```diff
@@ -0,0 +1,757 @@
...
+const TWENTY_EIGHT_BYTES: usize = 28; // needed for creating a u32
```
alexpArtos, line 14: This is also a strange comment, for similar reasons. A u32 is 4 bytes.

primitives/avn-common/src/event_types.rs
```diff
@@ -0,0 +1,757 @@
...
+const DISCARDED_ZERO_BYTES: usize = 12; // the first bytes of an Ethereum address are 0
```
alexpArtos, line 15:
```suggestion
const DISCARDED_ZERO_BYTES: usize = 12; // Used to ignore the first 12 bytes of a 32-byte value when creating an Ethereum address.
```

primitives/avn-common/src/event_types.rs
```diff
@@ -0,0 +1,757 @@
...
+            topics[Self::TOPIC_INDEX_SALE_INDEX][TWENTY_FOUR_BYTES..WORD_LENGTH]
+                .try_into()
+                .map_err(|_| Error::NftMintedEventSaleIndexConversion)?,
+        );
+        let t2_owner_public_key = H256::from_slice(&topics[Self::TOPIC_INDEX_T2_OWNER_PUBLIC_KEY]);
+
+        // This is a string field but its value should always be 4 WORD_LENGTH.
+        // The first 2 WORD_LENGTH are encoding data.
+        // The actual unique ref is expected to be a UUID which is made up of 32bytes (WORD_LENGTH)
+        // + 4bytes for the dashes Example: b1dc0452-8b2f-78ec-7e80-167002d11678
```
alexpArtos, line 346:
```suggestion
        // + 4 bytes for the dashes. Example: b1dc0452-8b2f-78ec-7e80-167002d11678
```

primitives/avn-common/src/event_types.rs
```diff
@@ -0,0 +1,757 @@
...
+impl Default for CheckResult {
+    fn default() -> Self {
+        CheckResult::Unknown
+    }
+}
+
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, Debug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+// Note: strictly speaking, different contracts can have events with the same signature, which would
+// suggests that the contract should be part of the EventId.
```
alexpArtos, line 624:
```suggestion
// suggest that the contract should be part of the EventId.
```

primitives/avn-common/src/event_types.rs
alexpArtos, line 13:
```suggestion
const WORD_LENGTH_MINUS_EIGHT_BYTES: usize = 24; // needed for creating a u64
```

primitives/avn-common/src/event_types.rs
alexpArtos, line 14:
```suggestion
const WORD_LENGTH_MINUS_FOUR_BYTES: usize = 28; // needed for creating a u32
```

alexpArtos (approved):

primitives/avn-common/src/event_types.rs
nahuseyoum: Please see usage of that constant.

primitives/avn-common/src/event_types.rs
nahuseyoum: Please see usage of the constant.

thadouk (approved):

thadouk (approved):
As discussed before, for as long as the repo is not used to produced the node binary, and its used to host library crates, the cargo.lock shouldn't be checked in
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
I would prefer if we removed it and added it to gitignore.

primitives/avn-common/src/event_types.rs
alexpArtos: I did, and that is why I proposed a name change that IMO communicates the usage more effectively.

primitives/avn-common/src/tests/test_avt_growth_event_parsing.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2022 Aventus Systems (UK) Ltd.
+#[cfg(test)]
+use super::*;
+use crate::{avn_tests_helpers::ethereum_converters::*, event_types::AvtGrowthLiftedData};
+use byte_slice_cast::AsByteSlice;
+use sp_core::U256;
+use sp_std::vec::Vec;
+
+pub fn into_4_be_bytes(bytes: &[u8]) -> Vec<u8> {
```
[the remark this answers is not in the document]
nahuseyoum: This is only used here for now. if / when they are used somewhere else we can centralise them.

primitives/avn-common/src/event_types.rs
nahuseyoum: > As discussed before, for as long as the repo is not used to produced the node binary, and its used to host library crates, the cargo.lock shouldn't be checked in https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries I would prefer if we removed it and added it to gitignore.

Ok, we will review the setup of the repo and the usage of that file and decide later.

PR #40, opened 2022-11-01
Adds tab_spaces to rustfmt.toml

thadouk:
Sets tab_spaces to 4 to match the original repo & team code style. https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#tab_spaces

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,5 +1,6 @@
 # Basic
 hard_tabs = false
+tab_spaces = 4
 max_width = 100
 use_small_heuristics = "Max"
 


alexpArtos (approved):

.rustfmt.toml
```diff
@@ -1,5 +1,6 @@
 # Basic
 hard_tabs = false
+tab_spaces = 4
```
nahuseyoum, line 3: Based on the rustfmt docs, this is the default value. If so, do we need to specify it?

thadouk: In most cases I think omitting the defaults is the right way to go. For options that are part of our coding style, adding it makes it more visible and saves the need to check the documentation.

nahuseyoum (approved):

PR #43, opened 2022-11-02
SYS-2849 Make staking compatible with sessions

nahuseyoum:
The original `parachain staking` pallet is not compatible with the session pallet, and since we have to use the session pallet, this PRs makes it compatible.

I am also removing references to `collator_selection` because the staking pallet will do that for us.


primitives/avn-common/src/tests/helpers.rs
```diff
@@ -10,6 +10,7 @@ pub mod ethereum_converters {
         return vec
     }
 
+    #[cfg(test)]
```
alexpArtos, line 13: This is a minor thing, but do we need this here? If this is a test helper, maybe we should only qualify with this directive the inclusion of the file itself in the callers? We wouldn't take part of the test helper without taking the rest, would we?

alexpArtos (approved):

fluorostani (approved):

pallets/parachain-staking/src/lib.rs
```diff
@@ -1160,6 +1143,44 @@ pub mod pallet {
...
     impl<T: Config> Pallet<T> {
+        pub fn start_new_era(
+            block_number: T::BlockNumber,
+            mut era: EraInfo<T::BlockNumber>,
+        ) -> (EraInfo<T::BlockNumber>, Weight) {
+            // mutate era
+            era.update(block_number);
+
+            // notify new era has begun
+            let mut weight = T::OnNewEra::on_new_era(era.current);
```
thadouk, line 1154: the original code was adding this to the weight of base_on_initialize, where the new code with the function discards it.
Is this intended?

thadouk (approved):
The PR was also doing a refactor of the tests, and mocks which increased the size of the PR significantly. I think it could have been broken in two.
Approved, but can we clarify that weight add/overwrite comment please?

pallets/parachain-staking/src/lib.rs
nahuseyoum: No it wasn't. Updated to retain the original weight.

PR #46, opened 2022-11-04
SYS-2851 Combines all "delay" values and makes it a storage item

nahuseyoum:
This PR combines the following config trait types into one value (`Delay`):
```
LeaveCandidatesDelay
CandidateBondLessDelay
LeaveNominatorsDelay
RevokeNominationDelay
NominationBondLessDelay
RewardPaymentDelay
```
and uses a storage item to control the value.


pallets/parachain-staking/src/lib.rs
```diff
@@ -392,6 +377,11 @@ pub mod pallet {
         }
     }
 
+    #[pallet::storage]
+    #[pallet::getter(fn delay)]
+    /// Number of eras to wait before executing any staking action
+    pub type Delay<T: Config> = StorageValue<_, EraIndex, ValueQuery>;
```
thadouk, line 383: I think we are missing a TODO here, to set the value on existing chains via a storage upgrade

thadouk (approved):

pallets/parachain-staking/src/lib.rs
nahuseyoum: There is a separate ticket for that.

alexpArtos (approved):
I like this, looks much more streamlined.

fluorostani (approved):

PR #50, opened 2022-11-08
SYS-2825 Update mock and tests with new `AccountId` type

nahuseyoum:
This PR updates the tests to use the new type definition for `AccountId`

### DIFF
--- a/pallets/parachain-staking/src/test_staking_pot.rs
+++ b/pallets/parachain-staking/src/test_staking_pot.rs
@@ -4,7 +4,22 @@ use frame_support::traits::Currency;
 
 pub const ONE_TOKEN: u128 = 1_000000_000000_000000u128;
 pub const AMOUNT_100_TOKEN: u128 = 100 * ONE_TOKEN;
-pub const NON_COLLATOR_ACCOUNT_ID: u64 = 2u64;
+
+fn collator_1() -> AccountId {
+    return TestAccount::new(1u64).account_id();
+}
+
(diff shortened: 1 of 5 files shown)


pallets/parachain-staking/src/test_staking_pot.rs
```diff
@@ -16,18 +31,21 @@ fn get_total_balance_of_collators(collator_account_ids: &Vec<AccountId>) -> u128
 
 #[test]
 fn fee_is_added_to_pot() {
+    let collator_1 = collator_1();
+    let collator_3 = collator_3();
+
     ExtBuilder::default()
-        .with_balances(vec![(1, 20), (2, 40), (3, 20), (4, 20)])
-        .with_candidates(vec![(1, 20), (3, 20), (4, 20)])
+        .with_balances(vec![(collator_1, 20), (non_collator_account_id(), 40), (collator_3, 20), (collator_4(), 20)])
```
fluorostani, line 38: Any reason why some collators are immediate getters whilst some are set up front?

fluorostani (approved):

alexpArtos (approved):

pallets/parachain-staking/src/test_staking_pot.rs
nahuseyoum: Its based on usage. If an account is used in a lot of places, I use a variable. Otherwise I call the getter directly.

PR #52, opened 2022-11-08
SYS-2758: integration test changed

RuiSMagalhaes:
**WHAT**
- Integration test changed to fetch all pallets used on the public repo

**TESTS**
- tested locally but test on the CI needs to be done after merging

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -47,7 +47,7 @@ jobs:
               export BRANCH=${GITHUB_REF#refs/heads/}
               export REPO="https://github.com/${{ github.repository }}"
           fi
-          sed "s@$STRING_TO_CHANGE@\1\"$REPO\"\3\4\"$BRANCH\"\6@" */**/*.toml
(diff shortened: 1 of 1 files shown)


vukomir (approved):

.github/workflows/integration_test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - name: Task pre integration test
         env:
-          STRING_TO_CHANGE: '\(pallet-parachain-staking.*\)\(".*\)"\(.*branch.*\)\(".*"\)\(.*\)'
+          STRING_TO_CHANGE: '\(.*\)\("https://github.com/Aventus-Network-Services/avn-parachain"\)\(.*\)\(branch.*\)\(".*"\)\(.*\)'
```
nahuseyoum, line 35: just to be 100% safe, should we add the `git = ` prefix on the url? Something like `git\s*=\s*` maybe?

nahuseyoum (approved):

.github/workflows/integration_test.yml
RuiSMagalhaes: discussed offline and decided to go without these changes ☝️

PR #59, opened 2022-11-15
SYS-2824 forkless upgrade data migration

nahuseyoum:
This PR adds a state migration script to enable staking via forkless upgrades


pallets/parachain-staking/src/tests/test_growth.rs
```diff
@@ -540,6 +540,7 @@ mod growth_amount {
                     let amount = 400;
                     assert_ok!(ParachainStaking::payout_collators(amount, PERIOD_INDEX));
 
+                    assert_ok!(ParachainStaking::payout_collators(amount, PERIOD_INDEX));
```
fluorostani, line 543: Is this test supposed to be calling the payout twice like this?

fluorostani (approved):

pallets/parachain-staking/src/tests/test_growth.rs
nahuseyoum: No, it shouldn't. Removed.

PR #67, opened 2022-11-17
SYS-1640 move nft manager pallet to this repo

nahuseyoum:
This PR does not change any functionality in the pallet. This is a move and some formatting changes applied.


fluorostani (approved):

pallets/nft-manager/src/batch_nft.rs
```diff
@@ -0,0 +1,215 @@
+use crate::*;
```
thadouk, line 1: Missing copyright

thadouk (approved):
I think cargo.lock should be updated as well since we add a new pallet.

PR #77, opened 2022-11-19
SYS-2825 update bond extra logic and tests

nahuseyoum:
This PR does the following:
 - It separate nominator bond extra and collator bond extra signature contexts and
 - adds invalid signature tests


alexpArtos (approved):

pallets/parachain-staking/src/tests/bond_extra_tests.rs
```diff
@@ -353,6 +404,39 @@ mod proxy_signed_candidate_bond_extra {
                 });
         }
 
+        #[test]
+        fn proxy_proof_is_not_valid() {
```
fluorostani, line 408: Are we just using a bad nonce to cover all invalid proof cases?

fluorostani (approved):

pallets/parachain-staking/src/tests/bond_extra_tests.rs
nahuseyoum: I will add a test for each parameter that gets signed.

PR #80, opened 2022-11-19
SYS-2825 adds unbonding tests

nahuseyoum:
Unbonding Test results: 
[unbondTestsResults.txt](https://github.com/Aventus-Network-Services/avn-parachain/files/10059253/unbondTestsResults.txt)


fluorostani (approved):

pallets/parachain-staking/src/tests/schedule_unbond_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+            .execute_with(|| {
+                let nonce = ParachainStaking::proxy_nonce(staker.account_id);
+                let unbond_call = create_call_for_signed_schedule_nominator_unbond(
+                    &staker,
+                    nonce,
+                    AMOUNT_TO_UNBOND.with(|v| *v.borrow()),
+                );
+                assert_ok!(AvnProxy::proxy(Origin::signed(staker.relayer), unbond_call, None));
+
+                assert!(unbond_event_emitted(Staker::default().account_id));
```
alexpArtos, line 180: Is the logic of balance changes checked somewhere else? These tests only assert success and event emitted.

alexpArtos, line 180: Is the logic of balance changes checked somewhere else? These tests only assert success and event emitted.

alexpArtos (approved):

pallets/parachain-staking/src/tests/schedule_unbond_tests.rs
nahuseyoum: Yes, the proxy (signed) version and the direct version share common code to do the balance changes and this has already been tested by the original pallet code.

PR #112, opened 2022-12-20
SYS-3007 fix benchmark value on signed_nominate

nahuseyoum:
This PR fixes a bug where calling `signed_nominate` was exhausting the block weight.


thadouk (approved):

thadouk (approved):
Can we run and update the value as well to avoid multiple PRs for this?

nahuseyoum:
> Can we run and update the value as well to avoid multiple PRs for this?

The good thing about this is that it doesn't require a benchmark rerun. These were variables we pass in to the calculated benchmark function.

micaelffrancoAV (approved):

PR #114, opened 2023-01-31
Fixes upgrade path for staking pallet

thadouk:
Adds current version to the pallet and updates the pallet code to handle the different upgrade paths

Removes dead code and warnings from tests.

Jira items:
- SYS-3111

### DIFF
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -718,18 +718,14 @@ pub mod pallet {
                 total_balance: total_staked,
             });
 
-            log::info!(
+            // Set storage version
+            crate::migration::STORAGE_VERSION.put::<Pallet<T>>();
+            log::debug!(
                 "Staking storage chain/current storage version: {:?} / {:?}",
                 Pallet::<T>::on_chain_storage_version(),
                 Pallet::<T>::current_storage_version(),
             );
 
-            // Set storage version
-            log::info!(
-                "Setting staking storage to latest version: {:?}",
-                crate::migration::STORAGE_VERSION
-            );
-            crate::migration::STORAGE_VERSION.put::<Pallet<T>>();
         }
     }
 
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -133,6 +133,7 @@ pub mod pallet {
     pub use sp_std::{collections::btree_map::BTreeMap, prelude::*};
     /// Pallet for parachain staking
     #[pallet::pallet]
+    #[pallet::storage_version(crate::migration::STORAGE_VERSION)]
     #[pallet::without_storage_info]
     pub struct Pallet<T>(PhantomData<T>);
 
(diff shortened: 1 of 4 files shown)


pallets/parachain-staking/src/migration.rs
```diff
@@ -145,10 +145,19 @@ pub fn enable_staking<T: Config>() -> Weight {
...
+        let onchain = Pallet::<T>::on_chain_storage_version();
 
-        if onchain_version < 1 {
+        log::info!(
+            "💽 Running migration with current storage version {:?} / onchain {:?}",
+            current,
+            onchain
+        );
+
+        if current == 1 && onchain == 0 {
```
nahuseyoum, line 157: Do we need this? Isn't this covered by the existing logic? I think we only want to run the migration if the `onchain` version is less than 1 right? Do we care about the current version?

pallets/parachain-staking/src/lib.rs
```diff
@@ -716,6 +717,19 @@ pub mod pallet {
...
+                "Staking storage chain/current storage version: {:?} / {:?}",
+                Pallet::<T>::on_chain_storage_version(),
+                Pallet::<T>::current_storage_version(),
+            );
+
+            // Set storage version
+            log::info!(
+                "Setting staking storage to latest version: {:?}",
+                crate::migration::STORAGE_VERSION
+            );
```
nahuseyoum, line 731: Do we need this? We can query the value from state if we need to.

thadouk: I think its cleaner if we set the version of the state, when we build the genesis state.
The original issue we are experiencing is because this was missing.

pallets/parachain-staking/src/migration.rs
thadouk: Depends I suppose.
The migration should run when the current implementation of the pallet is version 1.
Once we have a version 2, we can make a judgement call if it should be run or not. Till then its safer keeping it under this constrain.
Substrate pallets have similar conditions on upgrade.

pallets/parachain-staking/src/lib.rs
thadouk: I updated the logs a bit, and reduced the level.

thadouk: I updated the logs a bit, and reduced the level.

pallets/parachain-staking/src/tests/schedule_revoke_nomination_tests.rs
```diff
@@ -705,7 +693,7 @@ mod proxy_signed_execute_revoke_all_nomination {
                     roll_to_era_begin((ParachainStaking::delay() + 1u32) as u64);
 
                     let nonce = ParachainStaking::proxy_nonce(staker.account_id);
-                    let bad_nominator = to_acc_id(2000u64);
+                    let _bad_nominator = to_acc_id(2000u64);
```
fluorostani, line 696: why not just remove this?

alexpArtos (approved):

fluorostani (approved):

pallets/parachain-staking/src/lib.rs
nahuseyoum: I agree about setting the version on genesis. I meant about the logs. I think we can remove the logs as they don't add much value.

pallets/parachain-staking/src/migration.rs
nahuseyoum: I don't think so, unless I am missing something. Regardless of the current implementation, I think the migration should run when the **onchain** version is < 1.

nahuseyoum (approved):

PR #125, opened 2023-02-06
SYS-2850 Makes avn-proxy transactional compatible

thadouk:
This PR reafactors avn-proxy pallet to be compatible with v0.9.28 or newer versions of substrate that make all extrinsics transactional.
Up to now, if an error occurred and the extrinsic got rejected, storage changes up to that point were persistent. Avn-proxy pallet was using this functionality to give the fees to the relayer account, when the inner call failed to complete.
In order to persist that behaviour, this PR makes changes on the interface and the behaviour of the system.
Proxied extrinsics that the inner call fails to execute for any reason will be accepted and an `InnerCallFailed` event will be emitted, with the original dispatch error. Fee charges will be persistent in storage.

If the proxied extrinsic fails validation, the behaviour remains the same as it was prior to this PR.

The PR also updates all test that were expecting the proxied extrinsic to fail if the inner call fails. Instead now we look for the event and the appropriate extrinsic info.

### DIFF
--- a/pallets/avn-proxy/src/lib.rs
+++ b/pallets/avn-proxy/src/lib.rs
@@ -81,7 +81,15 @@ pub mod pallet {
     #[pallet::event]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        CallDispatched { relayer: T::AccountId, hash: T::Hash },
+        CallDispatched {
+            relayer: T::AccountId,
+            hash: T::Hash,
+        },
+        InnerCallFailedDispatch {
+            relayer: T::AccountId,
+            hash: T::Hash,
+            dispatch_error: DispatchError,
+        },
     }
 
     #[pallet::error]
@@ -126,8 +134,19 @@ pub mod pallet {
             let call_hash: T::Hash = T::Hashing::hash_of(&call);
             let sender: T::Origin = frame_system::RawOrigin::Signed(proof.signer.clone()).into();
 
-            call.dispatch(sender).map(|_| ()).map_err(|e| e.error)?;
-            Self::deposit_event(Event::<T>::CallDispatched { relayer, hash: call_hash });
+            let dispatch_result = call.dispatch(sender).map(|_| ()).map_err(|e| e.error);
+            match dispatch_result {
+                Ok(_) => {
+                    Self::deposit_event(Event::<T>::CallDispatched { relayer, hash: call_hash });
+                },
+                Err(dispatch_error) => {
+                    Self::deposit_event(Event::<T>::InnerCallFailedDispatch {
+                        relayer,
+                        hash: call_hash,
+                        dispatch_error,
+                    });
+                },
+            }
 
             Ok(Some(final_weight).into())
         }
--- a/pallets/avn-proxy/src/lib.rs
+++ b/pallets/avn-proxy/src/lib.rs
@@ -81,15 +81,8 @@ pub mod pallet {
     #[pallet::event]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        CallDispatched {
-            relayer: T::AccountId,
-            hash: T::Hash,
-        },
-        InnerCallFailed {
-            relayer: T::AccountId,
-            hash: T::Hash,
-            dispatch_error: DispatchError,
-        },
+        CallDispatched { relayer: T::AccountId, hash: T::Hash },
+        InnerCallFailed { relayer: T::AccountId, hash: T::Hash, dispatch_error: DispatchError },
     }
 
     #[pallet::error]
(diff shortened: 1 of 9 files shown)


pallets/avn-proxy/src/lib.rs
```diff
@@ -126,8 +134,19 @@ pub mod pallet {
             let call_hash: T::Hash = T::Hashing::hash_of(&call);
             let sender: T::Origin = frame_system::RawOrigin::Signed(proof.signer.clone()).into();
 
-            call.dispatch(sender).map(|_| ()).map_err(|e| e.error)?;
-            Self::deposit_event(Event::<T>::CallDispatched { relayer, hash: call_hash });
+            let dispatch_result = call.dispatch(sender).map(|_| ()).map_err(|e| e.error);
+            match dispatch_result {
+                Ok(_) => {
+                    Self::deposit_event(Event::<T>::CallDispatched { relayer, hash: call_hash });
```
nahuseyoum, line 140: I would go with `DispatchFailed`. 

But more generally, should we go with 2 events or 1 event with a status field? I know we went with the 2 events approach for EthereumEvents but just asking in case 1 event is simpler to check.

alexpArtos: I would use `InnerCallDispatchFailed` or `InnerCallFailed`. I think the former is equivalent grammatically to what Nahu suggested, but could give more context. I am not sure we can Dispatch anything else other than an inner call, so I won't make a fuss between the two.

nahuseyoum: I don't have a strong preference either, its just that `innerCall` is a bit redundant because there is nothing else we are dispatching in this pallet.

alexpArtos (approved):

pallets/avn-proxy/src/lib.rs
thadouk: I think DispatchFailed is an overloaded term, hence saying having the Inner part.
If you are both happy with InnerCallFailed I can go with that.
Regarding using one or two, if we go with one event, we are changing also an existing interface, which I was trying to avoid. Then the error field will have to become an optional field.
I think having a separate one will be cleaner.

pallets/avn-proxy/src/tests/mock.rs
```diff
@@ -293,6 +293,21 @@ pub fn proxy_event_emitted(
     })
 }
 
+pub fn proxy_failed_inner_dispatch_event_emitted(
```
alexpArtos, line 296: This function name no longer matches the event name

ivan-cholakov (approved):

pallets/avn-proxy/src/tests/proxy_tests_with_fees.rs
```diff
@@ -120,16 +119,19 @@ mod charging_fees {
...
+                    inner_call,
+                    payment_authorisation
+                ));
 
-                // No success events emitted
+                assert_eq!(
+                    true,
+                    proxy_failed_inner_dispatch_event_emitted(
+                        context.relayer.account_id(),
+                        call_hash
```
alexpArtos, line 132: In the previous version, we had an error indicating the source of the failure. Do we lose that with the new event? Is there any way we could wrap it in the dispatch failed event, or emit both for example?

alexpArtos (approved):

pallets/avn-proxy/src/tests/mock.rs
thadouk: updated

pallets/avn-proxy/src/tests/proxy_tests_with_fees.rs
thadouk: We can keep it, but I intentionally dropped it.
What is more valuable i think in the context of avn-proxy is matching that the InnerCallFailed event has been emited, that contains the DispatchError information.
Makes the tests more generic and testing the proxy behaviour.
In the staking pallet we do the inverted matching.

nahuseyoum: I think the tests  would be more robust if they tested for the specific error case.

pallets/parachain-staking/src/tests/bond_extra_tests.rs
```diff
@@ -288,9 +290,16 @@ mod proxy_signed_bond_extra {
...
-                        Error::<Test>::UnauthorizedSignedBondExtraTransaction
+                    assert_ok!(AvnProxy::proxy(
+                        Origin::signed(staker.relayer),
+                        bond_extra_call,
+                        None
+                    ));
+                    assert_eq!(
+                        true,
+                        inner_call_failed_event_emitted(
+                            Error::<Test>::UnauthorizedSignedBondExtraTransaction.into()
```
nahuseyoum, line 301: :+1:

nahuseyoum (approved):

pallets/avn-proxy/src/tests/proxy_tests_with_fees.rs
thadouk: Yes @alexpArtos, exactly that.

PR #135, opened 2023-04-29
SYS-3285 Update payment nonce handling - HOTFIX (DO NOT MERGE)

nahuseyoum:
This branch is based on v1.0.5 **not** main, hence the conflicts.  

This PR allows for the fee to be charged even if the inner call signature is invalid. 

If the payer has signed (agreed to pay) the inner call signature (the inner call proof), even if that proof is invalid, the payer will be charged and the payment nonce will be incremented
If the payment signature itself is not valid: 
 - the payer will not be charged
 - the payment nonce is not incremented
 - the inner call is not dispatched

### DIFF
--- a/pallets/avn-proxy/src/lib.rs
+++ b/pallets/avn-proxy/src/lib.rs
@@ -119,16 +119,18 @@ pub mod pallet {
                     .saturating_add(call.get_dispatch_info().weight)
                     .saturating_add(50_000);
 
-                    // Always try to charge a fee, regardless of the outcome of execution.
-                    // If the payment signature is not valid, the nonce is not incremented and the transaction is rejected
-                    // This allows to keep the payment nonce in memory when sending multiple transactions back to back.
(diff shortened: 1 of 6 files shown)


ivan-cholakov (approved):

alexpArtos:
If I understood this correctly, this is the same as #136, except that it is for an earlier version of the Substrate code that does not have the latest upgrade. Is that correct?

thadouk (approved):
Approved, same comment as:
https://github.com/Aventus-Network-Services/avn-parachain/pull/136#pullrequestreview-1408831423

nahuseyoum:
> If I understood this correctly, this is the same as #136, except that it is for an earlier version of the Substrate code that does not have the latest upgrade. Is that correct?

That's correct. This is the hotfix and the other is the regular PR to main.

nahuseyoum:
Release created and deployed on public testnet

PR #136, opened 2023-04-30
SYS-3285 Update payment nonce handling

nahuseyoum:
This PR allows for the fee to be charged even if the inner call signature is invalid.

If the payer has signed (agreed to pay) the inner call signature (the inner call proof), even if that proof is invalid, the payer will be charged and the payment nonce will be incremented
If the payment signature itself is not valid:

the payer will not be charged
the payment nonce is not incremented
the inner call is not dispatched


thadouk (approved):
Approved.
As a general remark, I think the proxy function needs a bit of documentation, to define the expected behaviour. When this succeeds, when it fails, when it is supposed to succeed but fail internally when we should charge and when we should roll back.
It will also help with the review and maintenance of the code because the edge cases complicate the code flow a bit.

alexpArtos (approved):

PR #151, opened 2023-05-30
SYS-3360 - registerCollator

ivan-cholakov:
This is redo of a previous PR, this time appending a new struct instead of modifying the already present struct

### DIFF
--- a/pallets/ethereum-transactions/src/lib.rs
+++ b/pallets/ethereum-transactions/src/lib.rs
@@ -448,12 +448,9 @@ impl<T: Config> Pallet<T> {
     fn get_contract_address(transaction_type: &EthTransactionType) -> Option<H160> {
         return match transaction_type {
             EthTransactionType::PublishRoot(_) => Some(Self::get_publish_root_contract()),
-            EthTransactionType::DeregisterValidator(_) =>
-                Some(T::ValidatorManagerContractAddress::get()),
-            EthTransactionType::SlashValidator(_) =>
-                Some(T::ValidatorManagerContractAddress::get()),
-            EthTransactionType::ActivateValidator(_) =>
-                Some(T::ValidatorManagerContractAddress::get()),
+            EthTransactionType::DeregisterValidator(_) |
+            EthTransactionType::SlashValidator(_) |
+            EthTransactionType::ActivateValidator(_) |
             EthTransactionType::ActivateCollator(_) =>
                 Some(T::ValidatorManagerContractAddress::get()),
             _ => None,
--- a/pallets/ethereum-transactions/src/lib.rs
+++ b/pallets/ethereum-transactions/src/lib.rs
@@ -454,6 +454,8 @@ impl<T: Config> Pallet<T> {
                 Some(T::ValidatorManagerContractAddress::get()),
             EthTransactionType::ActivateValidator(_) =>
                 Some(T::ValidatorManagerContractAddress::get()),
+            EthTransactionType::ActivateCollator(_) =>
+                Some(T::ValidatorManagerContractAddress::get()),
             _ => None,
         }
     }
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -37,14 +37,15 @@ use pallet_avn::{
 };
 use pallet_ethereum_transactions::{
     ethereum_transaction::{
-        ActivateValidatorData, DeregisterValidatorData, EthAbiHelper, EthTransactionType,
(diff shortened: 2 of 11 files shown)


pallets/validators-manager/src/tests.rs
```diff
@@ -683,15 +688,16 @@ mod add_validator {
             let context = &AddValidatorContext::default();
 
             set_session_keys(&context.collator);
-            assert_ok!(register_validator(&context.collator, &context.validator_eth_public_key));
+            println!("HELP !!! {:?}", &context.collator_eth_public_key);
```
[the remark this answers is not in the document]
ivan-cholakov: forgot to cleanup sorry

pallets/ethereum-transactions/src/lib.rs
```diff
@@ -454,6 +454,8 @@ impl<T: Config> Pallet<T> {
                 Some(T::ValidatorManagerContractAddress::get()),
             EthTransactionType::ActivateValidator(_) =>
                 Some(T::ValidatorManagerContractAddress::get()),
+            EthTransactionType::ActivateCollator(_) =>
```
nahuseyoum, line 457: Maybe we should add a TODO in this function because we only have 1 contract address and most of the match branches return the same one.

thadouk (approved):
Approved. provisionally.
The upgrade path should be tested before merging.
Additionally, I think the use of EthTransactionType::ActivateValidator and ActivateValidatorData should be marked as deprecated, via a [#[deprecated]](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) attribute and/or some comments.

Branch also needs updating from main.

nahuseyoum (approved):

alexpArtos (approved):

pallets/ethereum-transactions/src/lib.rs
ivan-cholakov: Can you elaborate please? The above comment also refers to the same line of code.

pallets/ethereum-transactions/src/lib.rs
```diff
@@ -454,6 +454,8 @@ impl<T: Config> Pallet<T> {
                 Some(T::ValidatorManagerContractAddress::get()),
             EthTransactionType::ActivateValidator(_) =>
```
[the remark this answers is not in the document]
ivan-cholakov: I guess this is an `or` like in Typescript how you define union types

pallets/ethereum-transactions/src/lib.rs
nahuseyoum: In `get_contract_address`, all match conditions, except publicRoot return the same contract address so we could combine them. Even publishRoot, although it reads from a different storage item, is returning the same contract address so we can add a TODO to simplify the logic here.

pallets/validators-manager/src/lib.rs
```diff
@@ -706,7 +704,8 @@ impl<T: Config> Pallet<T> {
     /// Helper function to help us fail early if any of the data we need is not available for the
     /// registration & activation
     fn prepare_registration_data(
-        validator_id: &T::AccountId,
+        collator_eth_id: &ecdsa::Public,
```
fluorostani, line 707: I think we should use `collator_eth_public_key` here too

fluorostani (approved):

thadouk (commented):
- Needs cleanup from debug info
- The implementation of the signature generation should be a separate PR and ticket.

pallets/validators-manager/src/lib.rs
```diff
@@ -248,25 +250,21 @@ pub mod pallet {
...
 
             ensure!(
                 !validator_account_ids.contains(&collator_account_id),
                 Error::<T>::ValidatorAlreadyExists
             );
             ensure!(
-                !<EthereumPublicKeys<T>>::contains_key(&validator_eth_public_key),
-                Error::<T>::ValidatorAlreadyExists
+                !<EthereumPublicKeys<T>>::contains_key(&collator_eth_public_key),
+                Error::<T>::ValidatorEthKeyAlreadyExists
```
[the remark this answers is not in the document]
thadouk: Errors are enumerations and in the past, the append-only rule applied to them as it could break the way the client decodes things if you alter them.
Not sure if it is still applicable since the clients are querying the runtime for the chain-specific types.

thadouk (commented):
Registration, Deregistration and the updated mechanism for signature calculation should come in a feature branch, once the whole process works and unit tests have been updated.

PR #154, opened 2023-06-01
Sys 3333 deregister collator

ivan-cholakov:
This pr adjusts the interface of the deregisterValidator function that interfaces with the smart contract. 
Before you submit a PR, please check that:

  - What does it do?
- [ ] Documentation updated

✄ -----------------------------------------------------------------------------


pallets/ethereum-transactions/src/lib.rs
```diff
@@ -448,11 +448,10 @@ impl<T: Config> Pallet<T> {
...
             EthTransactionType::PublishRoot(_) => Some(Self::get_publish_root_contract()),
-            EthTransactionType::DeregisterValidator(_) =>
-                Some(T::ValidatorManagerContractAddress::get()),
-            EthTransactionType::SlashValidator(_) =>
-                Some(T::ValidatorManagerContractAddress::get()),
-            EthTransactionType::ActivateValidator(_) =>
+            EthTransactionType::DeregisterValidator(_) |
+            EthTransactionType::ActivateValidator(_) |
+            EthTransactionType::ActivateCollator(_) |
+            EthTransactionType::DeregisterCollator(_) =>
```
alexpArtos, line 454:
```suggestion
            EthTransactionType::ActivateValidator(_) |
            EthTransactionType::DeregisterValidator(_) |
            EthTransactionType::ActivateCollator(_) |
            EthTransactionType::DeregisterCollator(_) =>
```

alexpArtos (approved):

pallets/validators-manager/src/mock.rs
```diff
@@ -74,6 +86,15 @@ const MOCK_T2_PUBLIC_KEY_BYTES: [u8; 32] = [
     151, 4, 182,
 ];
 
+fn hex_to_u8_array(hex: &str) -> [u8; 32] {
```
alexpArtos, line 89: Is this copied from some place? Could it be in some common place? 
(More like a nit-picking. Since this is a mock, it may be harder to refactor it to a common place than to maintain it as we go).

alexpArtos (approved):

thadouk (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -98,31 +101,46 @@ impl DeregisterValidatorData {
...
+                inputs: vec![
+                    Param {
+                        name: String::from("_targetT2PublicKey"),
+                        kind: ParamType::FixedBytes(32),
+                    },
+                    Param {
+                        name: String::from("_targetT1PublicKey"),
+                        kind: ParamType::FixedBytes(64),
+                    },
+                ],
```
alexpArtos, line 127: This is the contract we have in T1:
https://github.com/Aventus-Network-Services/avn-bridge/blob/master/contracts/AVNBridge.sol#L381

The order of arguments is:
- T1 public key
- T2 public key
- txId
- confirmations

So this order here is wrong.

PR #155, opened 2023-06-02
Aligns confirmation generation for adding collator

ivan-cholakov:
Reimplements the code that calculates the confirmations

Adds helper function that creates confirmations for calls that have their
arguments concatenated and hashed via keccak 256.

Uses the function above, to calculate confirmations for activate
collator, matching the expected values in avn-bridge.

Adds logs capturing the data generated and hashed.

Related JIRA items:
- SYS-3360
- SYS-3387

### DIFF
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -206,6 +206,8 @@ impl<T: Config> Pallet<T> {
         let mut url = String::from("eth/sign/");
         url.push_str(data_to_sign);
 
+        info!(target: "avn-service", "avn-service sign request (ecdsa) for data {:?}", data_to_sign);
+
         let ecdsa_signature_utf8 = Self::get_data_from_service(url)?;
         let ecdsa_signature_bytes = core::str::from_utf8(&ecdsa_signature_utf8)
             .map_err(|_| Error::<T>::ErrorConvertingUtf8)?;
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -206,7 +206,7 @@ impl<T: Config> Pallet<T> {
         let mut url = String::from("eth/sign/");
         url.push_str(data_to_sign);
 
-        info!(target: "avn-service", "avn-service sign request (ecdsa) for data {:?}", data_to_sign);
+        info!(target: "avn-service", "avn-service sign request (ecdsa) for hex-encoded data {:?}", data_to_sign);
 
         let ecdsa_signature_utf8 = Self::get_data_from_service(url)?;
         let ecdsa_signature_bytes = core::str::from_utf8(&ecdsa_signature_utf8)
--- a/pallets/ethereum-transactions/src/ethereum_transaction.rs
+++ b/pallets/ethereum-transactions/src/ethereum_transaction.rs
@@ -40,7 +40,6 @@ impl EthTransactionType {
             EthTransactionType::DeregisterValidator(d) => Ok(d.to_abi()),
             EthTransactionType::SlashValidator(d) => Ok(d.to_abi()),
             EthTransactionType::ActivateCollator(d) => Ok(d.to_abi()),
-            EthTransactionType::ActivateValidator(_d) => Err(EthAbiError::InvalidData),
             _ => Err(EthAbiError::InvalidData),
         }
     }
@@ -323,6 +322,19 @@ impl EthAbiHelper {
 
     pub fn generate_eth_abi_encoding_for_params_only(call: &EthTransactionDescription) -> Vec<u8> {
         return ethabi::encode(&call.call_values)
+    }
+
+    pub fn generate_confirmation_data_for_compacted_calls(
+        keccak_256_hash: &[u8; 32],
+        transaction_id: TransactionId,
+        from: &[u8; 32],
+    ) -> Vec<u8> {
+        let call_values: Vec<Token> = vec![
+            Token::FixedBytes(keccak_256_hash.to_vec()),
+            Token::Uint(EthAbiHelper::u256_to_big_endian(&U256::from(transaction_id)).into()),
+            Token::FixedBytes(from.to_vec()),
+        ];
+        return ethabi::encode(&call_values)
     }
 
     pub fn generate_ethereum_transaction_abi(
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -347,7 +347,10 @@ pub mod pallet {
         ) -> DispatchResult {
             ensure_none(origin)?;
 
-            let eth_encoded_data = Self::convert_data_to_eth_compatible_encoding(&action_id)?;
+            let eth_encoded_data =
+                Self::convert_collator_registration_action_to_confirmation_signing_data(
+                    &action_id, false,
+                )?;
             if !AVN::<T>::eth_signature_is_valid(eth_encoded_data, &validator, &approval_signature)
             {
                 create_and_report_validators_offence::<T>(
@@ -474,7 +477,10 @@ pub mod pallet {
                 }
 
                 let voting_session = Self::get_voting_session(action_id);
-                let eth_encoded_data = Self::convert_data_to_eth_compatible_encoding(action_id)
+                let eth_encoded_data =
+                    Self::convert_collator_registration_action_to_confirmation_signing_data(
+                        action_id, false,
+                    )
                     .map_err(|_| {
                         InvalidTransaction::Custom(ERROR_CODE_INVALID_DEREGISTERED_VALIDATOR)
                     })?;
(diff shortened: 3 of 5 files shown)


pallets/validators-manager/src/lib.rs
```diff
@@ -621,6 +623,25 @@ impl<T: Config> Pallet<T> {
...
+    pub fn convert_data_to_eth_compatible_encoding_for_collator_registration(
+        action_id: &ActionId<T::AccountId>,
+    ) -> Result<String, DispatchError> {
+        let validators_action_data = Self::try_get_validators_action_data(action_id)?;
+        let eth_description = EthAbiHelper::generate_ethereum_description_for_signature_request(
+            &T::AccountToBytesConvert::into_bytes(&validators_action_data.primary_validator),
+            &validators_action_data.reserved_eth_transaction,
+            validators_action_data.eth_transaction_id,
+        )
+        .map_err(|_| Error::<T>::ErrorGeneratingEthDescription)?;
```
alexpArtos, line 635: This is the bit we want to remove. Instead, create the array directly for the call to `encode(keccak(...))`

thadouk (approved):

pallets/avn/src/lib.rs
```diff
@@ -206,6 +206,8 @@ impl<T: Config> Pallet<T> {
         let mut url = String::from("eth/sign/");
         url.push_str(data_to_sign);
 
+        info!(target: "avn-service", "avn-service sign request (ecdsa) for data {:?}", data_to_sign);
```
alexpArtos, line 209: Please, add the hex version of the encoded data if possible. That is the definitive information we need to have, not the raw human-readable data (that is useful for debugging, ofc, but it's the encoding that matters)

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -325,6 +324,19 @@ impl EthAbiHelper {
         return ethabi::encode(&call.call_values)
     }
 
+    pub fn generate_confirmation_data_for_compacted_calls(
+        keccak_256_hash: &[u8; 32],
+        transaction_id: TransactionId,
+        from: &[u8; 32],
+    ) -> Vec<u8> {
+        let call_values: Vec<Token> = vec![
+            Token::FixedBytes(keccak_256_hash.to_vec()),
```
alexpArtos, line 333: Not sure if the root (in a publishRoot call) is a keccak hash. If so, I like this name. If not, it might be misleading, and I'd call it just `hash_256`.

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -325,6 +324,19 @@ impl EthAbiHelper {
         return ethabi::encode(&call.call_values)
     }
 
+    pub fn generate_confirmation_data_for_compacted_calls(
```
nahuseyoum, line 327: I can't see how the function name matches what it does. There is no confirmation logic here, as far as I can see, this function is abi encoding call data parameters.

pallets/validators-manager/src/lib.rs
```diff
@@ -587,6 +594,7 @@ const NAME: &'static [u8; 17] = b"validatorsManager";
 
 // Error codes returned by validate unsigned methods
 const ERROR_CODE_INVALID_DEREGISTERED_VALIDATOR: u8 = 10;
+const PACKED_ARGUMENTS_SIZE: usize = 96;
```
alexpArtos, line 597:
```suggestion
const PACKED_KEYS_SIZE: usize = 96;
```

pallets/validators-manager/src/lib.rs
```diff
@@ -619,6 +630,58 @@ impl<T: Config> Pallet<T> {
...
 
+    /// This function generates the data needed to generate a confirmation for registering a new
+    /// collator. The implementation must match this schema:
+    /// https://github.com/Aventus-Network-Services/avn-bridge/blob/v1.1.0/contracts/AVNBridge.sol#L344-L345
+    pub fn convert_collator_registration_action_to_confirmation_signing_data(
+        action_id: &ActionId<T::AccountId>,
+        enable_data_logs: bool,
+    ) -> Result<String, DispatchError> {
+        let validators_action_data = Self::try_get_validators_action_data(action_id)?;
+        let mut activate_collator_params_concat: [u8; PACKED_ARGUMENTS_SIZE] = [0u8; 96];
```
alexpArtos, line 641:
```suggestion
        let mut activate_collator_params_concat: [u8; PACKED_KEYS_SIZE] = [0u8; 96];
```
Is it possible to use the constant instead of `96` on the right-hand side as well?

pallets/validators-manager/src/lib.rs
```diff
@@ -619,6 +630,58 @@ impl<T: Config> Pallet<T> {
         Ok(hex::encode(EthAbiHelper::generate_eth_abi_encoding_for_params_only(&eth_description)))
     }
 
+    /// This function generates the data needed to generate a confirmation for registering a new
+    /// collator. The implementation must match this schema:
+    /// https://github.com/Aventus-Network-Services/avn-bridge/blob/v1.1.0/contracts/AVNBridge.sol#L344-L345
+    pub fn convert_collator_registration_action_to_confirmation_signing_data(
+        action_id: &ActionId<T::AccountId>,
+        enable_data_logs: bool,
```
nahuseyoum, line 638: I am not sure about this. I think we should have a better way to handle enabling/disabling logs.

pallets/validators-manager/src/lib.rs
nahuseyoum, line 641:
```suggestion
        let mut activate_collator_params_concat: [u8; PACKED_ARGUMENTS_SIZE] = [0u8; PACKED_ARGUMENTS_SIZE];
```

alexpArtos (approved):

pallets/validators-manager/src/lib.rs
```diff
@@ -619,6 +630,58 @@ impl<T: Config> Pallet<T> {
...
+            EthTransactionType::ActivateCollator(ref d) => d,
+            _ => Err(Error::<T>::ErrorGeneratingEthDescription)?,
+        };
+
+        activate_collator_params_concat[0..64]
+            .copy_from_slice(&activate_collator_params.t1_public_key.as_bytes()[0..64]);
+        activate_collator_params_concat[64..96]
+            .copy_from_slice(&activate_collator_params.t2_public_key[0..32]);
+
+        let activate_collator_hash = keccak_256(&activate_collator_params_concat);
```
nahuseyoum, line 653: We can extract this out into a separate function. It would make it easier to read and follow.

pallets/validators-manager/src/lib.rs
```diff
@@ -619,6 +630,58 @@ impl<T: Config> Pallet<T> {
         Ok(hex::encode(EthAbiHelper::generate_eth_abi_encoding_for_params_only(&eth_description)))
     }
 
+    /// This function generates the data needed to generate a confirmation for registering a new
+    /// collator. The implementation must match this schema:
+    /// https://github.com/Aventus-Network-Services/avn-bridge/blob/v1.1.0/contracts/AVNBridge.sol#L344-L345
+    pub fn convert_collator_registration_action_to_confirmation_signing_data(
```
nahuseyoum, line 636:
```suggestion
    pub fn abi_encode_collator_registration_data(
```

pallets/validators-manager/src/lib.rs
```diff
@@ -619,6 +630,58 @@ impl<T: Config> Pallet<T> {
...
+        // Now treat this as an bytes32 parameter and generate signing abi.
+        let hex_encoded_confirmation_data =
+            hex::encode(EthAbiHelper::generate_confirmation_data_for_compacted_calls(
+                &activate_collator_hash,
+                validators_action_data.eth_transaction_id,
+                &sender,
+            ));
+
+        if enable_data_logs {
+            log::info!(
```
nahuseyoum, line 667: we can also move this into the separate function

pallets/ethereum-transactions/src/ethereum_transaction.rs
thadouk: For now, this is for "compacted" calls and the parameter has a keccak_ prefix to make it obvious.
If/when we consume it in summary we should make it more generic and names more friendly.
Can do it now if you prefer that, and probably update the name as well

pallets/avn/src/lib.rs
thadouk: it is the hex encoded data. I will add it to the log to make it more obvious.

thadouk (approved):

pallets/validators-manager/src/lib.rs
thadouk: I don't think using a function will improve readability.

thadouk (approved):

fluorostani (approved):

pallets/validators-manager/src/lib.rs
nahuseyoum: I think we should remove the flag and always have the logs for now at the very least

pallets/validators-manager/src/lib.rs
nahuseyoum: I think it would significantly improve the readability

alexpArtos: If the function is well named, it should. It always should.

PR #156, opened 2023-06-08
SYS-3333 deregistration tests

thadouk:
This reverts commit 056a068d3cf00af9bd7e0ec8b4607974249a94bb.

# Thank you for your Pull Request

AFTER YOU'VE READ THIS NOTICE, FEEL FREE TO REMOVE IT.
Thank you!

Before you submit a PR, please check that:

  - What does it do?
- [ ] Documentation updated

✄ -----------------------------------------------------------------------------

### DIFF
--- a/pallets/ethereum-transactions/src/ethereum_transaction.rs
+++ b/pallets/ethereum-transactions/src/ethereum_transaction.rs
@@ -117,20 +117,20 @@ impl DeregisterCollatorData {
                 name: String::from("deregisterValidator"),
                 inputs: vec![
                     Param {
-                        name: String::from("_targetT2PublicKey"),
-                        kind: ParamType::FixedBytes(32),
+                        name: String::from("_targetT1PublicKey"),
+                        kind: ParamType::Bytes,
                     },
                     Param {
-                        name: String::from("_targetT1PublicKey"),
-                        kind: ParamType::FixedBytes(64),
+                        name: String::from("_targetT2PublicKey"),
+                        kind: ParamType::FixedBytes(32),
                     },
                 ],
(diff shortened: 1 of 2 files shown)


pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
```diff
@@ -28,21 +33,24 @@ fn generate_publish_root_eth_txn_desc(root_hash: [u8; 32]) -> EthTransactionDesc
...
         function_call: Function {
             name: String::from("deregisterValidator"),
-            inputs: vec![Param {
-                name: String::from("_targetT2PublicKey"),
-                kind: ParamType::FixedBytes(32),
-            }],
+            inputs: vec![
+                Param { name: String::from("_targetT2PublicKey"), kind: ParamType::FixedBytes(32) },
+                Param { name: String::from("_targetT1PublicKey"), kind: ParamType::FixedBytes(64) },
+            ],
```
alexpArtos, line 46: I think this is wrong. The call in the contract is this:
`  function deregisterValidator(bytes memory t1PublicKey, bytes32 t2PublicKey, uint256 t2TransactionId,
      bytes calldata confirmations)`
(https://github.com/Aventus-Network-Services/avn-bridge/blob/master/contracts/AVNBridge.sol#L345)

So, the first argument is the T1 public key, and the second is the T2 public key. These arguments are reversed in here.

True, the hash is computed the other way around, but that does not have an impact in the call.

thadouk:
PR will be re-opened on a clean branch, once tests have been updated

PR #157, opened 2023-06-08
SYS-3387 Adds tests for new confirmation schema

thadouk:
Splits `concat_and_hash_activation_data` and `concat_and_hash_deregistration_data` into a separate module and adds unit tests that ensure that the data that will be used for the confirmation signature are as expected.


alexpArtos (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -98,31 +101,46 @@ impl DeregisterValidatorData {
...
+                inputs: vec![
+                    Param {
+                        name: String::from("_targetT2PublicKey"),
+                        kind: ParamType::FixedBytes(32),
+                    },
+                    Param {
+                        name: String::from("_targetT1PublicKey"),
+                        kind: ParamType::FixedBytes(64),
+                    },
+                ],
```
alexpArtos, line 127: As mentioned in other PRs, this has to be changed. The inputs in the contract  are T1 key first, and then T2 key.

https://github.com/Aventus-Network-Services/avn-bridge/blob/master/contracts/AVNBridge.sol#L381

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
```diff
@@ -28,21 +33,24 @@ fn generate_publish_root_eth_txn_desc(root_hash: [u8; 32]) -> EthTransactionDesc
...
+                Param { name: String::from("_targetT1PublicKey"), kind: ParamType::FixedBytes(64) },
+            ],
             outputs: Vec::<Param>::new(),
             constant: false,
         },
-        call_values: vec![Token::FixedBytes(t2_public_key.to_vec())],
+        call_values: vec![
+            Token::FixedBytes(t2_public_key.to_vec()),
+            Token::Bytes(t1_public_key.to_fixed_bytes().to_vec()),
+        ],
```
alexpArtos, line 53: The order needs to be changed. T1 key comes first than T2:
https://github.com/Aventus-Network-Services/avn-bridge/blob/master/contracts/AVNBridge.sol#L381

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
```diff
@@ -81,10 +95,14 @@ mod eth_transaction_type {
...
-                let deregister_validator_eth_txn_desc =
-                    generate_deregister_validator_eth_txn_desc(T2_PUBLIC_KEY);
+                let deregister_validator_eth_txn_type = generate_deregister_validator_eth_txn_type(
+                    H512::from(T1_PUBLIC_KEY),
+                    T2_PUBLIC_KEY,
+                );
+                let deregister_validator_eth_txn_desc = generate_deregister_validator_eth_txn_desc(
+                    H512::from(T1_PUBLIC_KEY),
+                    T2_PUBLIC_KEY,
+                );
```
alexpArtos, line 105: I think we have to review this style of writing tests. I don't think it serves a point. What we are doing here is:
- mimicking the exact code from the pallet
- making sure that the tests agree with what the code has

This does not sound right to me. We are creating two parallel code bases and ensuring they do the same thing. Why? If we did the code wrong, what would convince us that the same algorithm and options would be right in the tests? It would still be as wrong!

So, what should we be testing instead? Functionality, behaviour. In this case, external verification perhaps: we need a transaction to be well-encoded enough that it will be accepted by Ethereum. We cannot do an integration test at this level, obviously, but that does not mean we have to do extensive tests of every function  (in particular, those that are not part of the public interface) if we don't have to. So, perhaps a better way here would be to pick as expected encoding a valid input that we know is accepted by Ethereum. Say we produced it outside and tested it outside. Then, our test would invoke the actual code and try to match the outside's encoding. 

That would be a smarter way of testing, in my opinion, that would reduce our test-writing time, be more meaningful (no white-box testing) and more maintainable (no need to change the tests when we refactor the implementation).

pallets/validators-manager/src/confirmations.rs
```diff
@@ -0,0 +1,109 @@
...
+    log::debug!(
+            "🗜️ Creating packed hash for {:?} transaction: Concat params data (hex encoded): {:?} - keccak_256 hash (hex encoded): {:?}",
+                &deregister_collator_data,
+                hex::encode(deregister_collator_params_concat),
+                hex::encode(deregister_collator_hash)
+        );
+    return deregister_collator_hash
+}
+
+// Tests are
```
alexpArtos, line 59: unfinished comment

pallets/validators-manager/src/confirmations.rs
```diff
@@ -0,0 +1,109 @@
...
+    let ferdie_activation_data = ActivateCollatorData {
+        t1_public_key: H512::from(ferdie_t1_public_key_bytes),
+        t2_public_key: ferdie_t2_public_key_bytes,
+    };
+
+    let hashed_keys = concat_and_hash_activation_data(&ferdie_activation_data);
+
+    assert_eq!(
+        hex::encode(&hashed_keys),
+        "fcde037cef635ab9da60f50efd8552403f7a7e6e58f1f1be3aba810ff99228ea"
```
alexpArtos, line 81: And this would be good with a comment as well. It is enough to say it was tested in Ethereum, if that was the case.

pallets/validators-manager/src/confirmations.rs
alexpArtos: Don't we usually write tests in separate files? Why is this in the same as the code?

pallets/ethereum-transactions/src/ethereum_transaction.rs
thadouk: This has been addressed in #165.

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -98,31 +101,46 @@ impl DeregisterValidatorData {
...
                 outputs: Vec::<Param>::new(),
                 constant: false,
             },
-            call_values: vec![Token::FixedBytes(self.t2_public_key.to_vec())],
+            call_values: vec![
+                Token::FixedBytes(self.t2_public_key.to_vec()),
+                Token::Bytes(self.t1_public_key.to_fixed_bytes().to_vec()),
+            ],
```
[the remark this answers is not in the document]
thadouk: This has been addressed in #154.

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
thadouk: This has been addressed in #165 .

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
thadouk: Comment migrated to #165 where it will be addressed:
https://github.com/Aventus-Network-Services/avn-parachain/pull/165#discussion_r1228777283

alexpArtos (approved):
It's acceptable. But it's still doing two parallel code bases and asserting they're equal.
The real proof is only when this is accepted in Ethereum, which has to be an integration test.
For the purpose of unit tests, I don't think we can do much better than this.

ivan-cholakov (approved):

PR #160, opened 2023-06-08
SYS-3357 Discount tests suite

micaelffrancoAV:
What: Tests that check the fees after different discounts are applied.

### DIFF
--- a/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
+++ b/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
@@ -144,12 +144,12 @@ mod adjustment_fee_tests {
                     assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
 
                     pay_gas_and_call_remark(&sender);
-                    assert_eq!(INITIAL_SENDER_BALANCE, Balances::free_balance(sender));
+                    assert_eq!(INITIAL_SENDER_BALANCE - u128::from(BASE_FEE), Balances::free_balance(sender));
 
                     assert!(event_emitted(&mock::RuntimeEvent::AvnTransactionPayment(
                         crate::Event::<TestRuntime>::AdjustedTransactionFeePaid {
                             who: sender,
-                            fee: 0u128, // sender got 100% refund
+                            fee: u128::from(BASE_FEE), // sender paid base fee
                         }
                     )));
                 })
--- a/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
+++ b/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
@@ -84,7 +84,7 @@ pub(crate) fn roll_blocks(n: u64) {
     }
 }
 
-mod discount_tests {
+mod adjustment_fee_tests {
     use super::*;
 
     mod fees_are_paid_correctly_when {
--- a/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
+++ b/pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
@@ -1,7 +1,7 @@
 use super::*;
 use crate::mock::{
-    event_emitted, new_test_ext, roll_one_block, AccountId, Balances, RuntimeCall, RuntimeEvent,
-    RuntimeOrigin, System, TestAccount, TestRuntime,
+    event_emitted, new_test_ext, roll_one_block, AccountId, AvnTransactionPayment, Balances,
+    RuntimeCall, RuntimeEvent, RuntimeOrigin, System, TestAccount, TestRuntime,
 };
 
 use frame_support::{dispatch::DispatchInfo, pallet_prelude::Weight};
@@ -12,9 +12,9 @@ use frame_support::assert_ok;
 
 pub const TX_LEN: usize = 1;
 pub const INITIAL_SENDER_BALANCE: u128 = 200;
-pub const BASE_FEE: u32 = 14;
+pub const BASE_FEE: u128 = 14;
 pub const FIXED_FEE: u128 = 10;
-pub const PERCENTAGE_FEE: u32 = 50;
+pub const PERCENTAGE_FEE: u32 = 25;
 
 fn to_acc_id(id: u64) -> AccountId {
     return TestAccount::new(id).account_id()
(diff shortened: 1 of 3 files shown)


pallets/avn-transaction-payment/src/tests/discount_fee_tests.rs
```diff
@@ -0,0 +1,460 @@
...
+                    assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
+
+                    pay_gas_and_call_remark(&sender);
+                    let percentage_value = get_percentage_fee_value(PERCENTAGE_FEE);
+                    assert_eq!(
+                        INITAL_SENDER_BALANCE - u128::from(percentage_value),
+                        Balances::free_balance(sender)
+                    );
+
+                    roll_one_block();
```
alexpArtos, line 370: Perhaps we could have a higher duration test as well? 3 blocks?

nahuseyoum (approved):

pallets/avn-transaction-payment/src/tests/discount_fee_tests.rs
```diff
@@ -0,0 +1,460 @@
...
+                    let percentage_value = get_percentage_fee_value(PERCENTAGE_FEE);
+                    assert_eq!(
+                        INITAL_SENDER_BALANCE - u128::from(percentage_value),
+                        Balances::free_balance(sender)
+                    );
+                })
+            }
+
+            #[test]
+            fn and_percentage_fee_is_more_than_100() {
```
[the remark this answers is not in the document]
micaelffrancoAV: This has been fixed, we now have a test for a 100% case which refunds the total amount.
Anything over 100 will fail and the call will error.

pallets/avn-transaction-payment/src/tests/discount_fee_tests.rs
```diff
@@ -0,0 +1,460 @@
...
+                    assert_ok!(AvnTransactionPayment::set_known_sender(
+                        RuntimeOrigin::root(),
+                        sender,
+                        config,
+                    ));
+                    assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
+
+                    pay_gas_and_call_remark(&sender);
+                    assert_eq!(
+                        INITAL_SENDER_BALANCE - u128::from(BASE_FEE),
```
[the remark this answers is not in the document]
micaelffrancoAV: When we provide a fixed fee higher than the base fee, the expected behaviour is to pay the base_fee. 
Users won't pay more than the base fee.

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
```diff
@@ -0,0 +1,582 @@
...
+use pallet_transaction_payment::ChargeTransactionPayment;
+use sp_runtime::traits::SignedExtension;
+
+use frame_support::assert_ok;
+
+pub const TX_LEN: usize = 1;
+pub const INITIAL_SENDER_BALANCE: u128 = 200;
+pub const BASE_FEE: u32 = 14;
+pub const FIXED_FEE: u128 = 10;
+pub const PERCENTAGE_FEE: u32 = 50;
```
alexpArtos, line 17: Let's use something different please. 30 for example. What I want to see is: the discount rate is 30%, the fee we paid is 70% of the original. 50% does not allow us to disambiguate this.

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
```diff
@@ -0,0 +1,582 @@
...
+    System::inc_account_nonce(sender);
+}
+
+fn set_initial_sender_balance(sender: &AccountId) {
+    Balances::make_free_balance_be(&sender, INITIAL_SENDER_BALANCE);
+    assert_eq!(INITIAL_SENDER_BALANCE, Balances::free_balance(sender));
+}
+
+fn get_percentage_fee_value(percentage_fee: u32) -> u32 {
+    (BASE_FEE * percentage_fee) / 100
```
alexpArtos, line 59: Shouldn't there be a test here for when this rate is invalid, that is, outside the [0;1] range?
It should not be negative or larger than 1, otherwise we'll get very odd behaviour.

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
```diff
@@ -0,0 +1,582 @@
...
+
+                    assert_ok!(AvnTransactionPayment::set_known_sender(
+                        RuntimeOrigin::root(),
+                        sender,
+                        config,
+                    ));
+                    assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
+
+                    pay_gas_and_call_remark(&sender);
+                    assert_eq!(INITIAL_SENDER_BALANCE - u128::from(BASE_FEE), Balances::free_balance(sender));
```
alexpArtos, line 147: I'm confused here. So what is the config in this case? Is it a discount whose value is the `base_fee + 1`? But then, shouldn't we have not paid `0`, instead of paying `base_fee`?

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
```diff
@@ -0,0 +1,582 @@
...
+                    assert_ok!(AvnTransactionPayment::set_known_sender(
+                        RuntimeOrigin::root(),
+                        sender,
+                        config,
+                    ));
+                    assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
+
+                    pay_gas_and_call_remark(&sender);
+
+                    let percentage_value = get_percentage_fee_value(PERCENTAGE_FEE);
```
alexpArtos, line 181: I think we should add another test condition here, to make it very clear what is the base of the percentage. Is that `base_fee`? So, is the fee we paid, say, `(1 - discount rate) * BASE_FEE`?

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
```diff
@@ -0,0 +1,582 @@
...
+                    };
+
+                    assert_ok!(AvnTransactionPayment::set_known_sender(
+                        RuntimeOrigin::root(),
+                        sender,
+                        config,
+                    ));
+                    assert_eq!(AvnTransactionPayment::is_known_sender(sender), true);
+
+                    pay_gas_and_call_remark(&sender);
```
alexpArtos, line 256: It would be so much easier to review the tests if all of this were captured into a single function called `check_sender_paid_fee(sender, fee)` making sure that the event and the balance change are in sync.

Also, removing the difference in abstraction level. When reading the code at this level, I can see:
- prepare balance
- make payment
- check fees
- prepare balance
- make payment
- check different fees (if the fee conditions have expired, for example)

This is not a request for change, now, but just an awareness example of how it could save time in reviewing (and in debugging tests six months down the line) with just a small approach to how to write tests.

alexpArtos (approved):

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
micaelffrancoAV: There are tests for that in the set_know_sender_tests suite!

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
micaelffrancoAV: The value we pass is not a discount, it is a fixed fee. Meaning the value we pass in will be the value we pay for the fee.

If we provide the fixed fee of 15, we don't charge anything more that the chain base fee. 
We just do a discount if the fixed fee is lower than our base fee.

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
micaelffrancoAV: I changed the `get_percentage_fee_value` to `get_percentage_fee_paid` and now returns the value paid with the percentage given.

pallets/avn-transaction-payment/src/tests/adjustment_fee_tests.rs
micaelffrancoAV: I agree, the tests have been refactored with the `check_sender_paid_fee` function and some more function to make it easier to read.

PR #164, opened 2023-06-12
updated benchmarking for validators-manager pallet

ivan-cholakov:
Updated validators-manager `benchmarks.rs` so that tests compile

PR #165, opened 2023-06-13
SYS-3333 Alignment of unit tests to feature - part 1

thadouk:
This PR:
- Fixes ethereum-transaction tests, by reversion the call order of validator deregistration
- Removes not used code and updates tests function to pass compilation on validator manager pallet, in alignment with the feature implementation. As is the test cases will fail but compile successfully. The test cases logic will be updated and refactored in follow up part of the changes.

This commit re-instates the revert commit https://github.com/Aventus-Network-Services/avn-parachain/commit/056a068d3cf00af9bd7e0ec8b4607974249a94bb which was split originally from #155 to reduce its scope.

This PR is a pre-requisitive for #164


pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
```diff
@@ -81,10 +95,14 @@ mod eth_transaction_type {
...
-                let deregister_validator_eth_txn_desc =
-                    generate_deregister_validator_eth_txn_desc(T2_PUBLIC_KEY);
+                let deregister_validator_eth_txn_type = generate_deregister_collator_eth_txn_type(
+                    H512::from(T1_PUBLIC_KEY),
+                    T2_PUBLIC_KEY,
+                );
+                let deregister_validator_eth_txn_desc = generate_deregister_collator_eth_txn_desc(
+                    H512::from(T1_PUBLIC_KEY),
+                    T2_PUBLIC_KEY,
+                );
```
thadouk, line 105: Copying review comment from: https://github.com/Aventus-Network-Services/avn-parachain/pull/157#discussion_r1227847028
> I think we have to review this style of writing tests. I don't think it serves a point. What we are doing here is:
> 
> * mimicking the exact code from the pallet
> * making sure that the tests agree with what the code has
> 
> This does not sound right to me. We are creating two parallel code bases and ensuring they do the same thing. Why? If we did the code wrong, what would convince us that the same algorithm and options would be right in the tests? It would still be as wrong!
> 
> So, what should we be testing instead? Functionality, behaviour. In this case, external verification perhaps: we need a transaction to be well-encoded enough that it will be accepted by Ethereum. We cannot do an integration test at this level, obviously, but that does not mean we have to do extensive tests of every function (in particular, those that are not part of the public interface) if we don't have to. So, perhaps a better way here would be to pick as expected encoding a valid input that we know is accepted by Ethereum. Say we produced it outside and tested it outside. Then, our test would invoke the actual code and try to match the outside's encoding.
> 
> That would be a smarter way of testing, in my opinion, that would reduce our test-writing time, be more meaningful (no white-box testing) and more maintainable (no need to change the tests when we refactor the implementation).

alexpArtos (approved):

ivan-cholakov (approved):

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
thadouk: The intention is not to do an extensive unit test using the same code.
The test is a failsafe in case code changes or refactors cause a change in the behaviour of the output of the function.

But I agree with what you said, especially with the test code mimicking the code of the pallet.
If it is ok I would like to address this on a separate PR so the refactor is separated from the validators work

PR #167, opened 2023-06-14
SYS-3402 fix duplicate AVT balance recording

nahuseyoum:
Fixes duplicate balance recording when performing AVT lifts and updates tests

### DIFF
--- a/pallets/token-manager/src/lib.rs
+++ b/pallets/token-manager/src/lib.rs
@@ -689,14 +689,16 @@ impl<T: Config> Pallet<T> {
                 amount: updated_amount,
                 eth_tx_hash: event_id.transaction_hash,
             });
+        } else {
+            Self::update_token_balance(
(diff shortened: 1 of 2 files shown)


ivan-cholakov (approved):

pallets/token-manager/src/lib.rs
```diff
@@ -689,14 +689,16 @@ impl<T: Config> Pallet<T> {
...
+            )?;
         }
 
-        return Self::update_token_balance(
-            event_id.transaction_hash,
-            data.token_contract.into(),
-            recipient_account_id,
-            data.amount,
-        )
+        return Ok(())
```
ivan-cholakov, line 701: would this occur because of `fmt`?

nahuseyoum: No, its human error unfortunately. During refactor we incorrectly removed the early return statement.

PR #168, opened 2023-06-14
SYS-3333: Update and fix unit tests - Part 2

thadouk:
This commit aligns the unit tests with the new implementation. It includes the following updates and fixes:

- Updates the mocks to use valid ECDSA signatures as confirmations.
- Modifies the test accounts to be derived from test private keys with a pattern of [1u8;32].
- Corrects internal logic in the tests that previously used the same AuthorityID for all validators.
- Removes tests that were specifically designed to test a mock implementation.

This PR rectifies any failing tests in #165

### DIFF
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -252,7 +252,8 @@ impl<T: Config> Pallet<T> {
         match T::EthereumPublicKeyChecker::get_validator_for_eth_public_key(
             &recovered_public_key.expect("Checked for error"),
         ) {
-            Some(maybe_validator) => maybe_validator == validator.account_id,
+            Some(maybe_validator) => {
+                maybe_validator == validator.account_id},
             _ => false,
         }
     }
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -252,8 +252,7 @@ impl<T: Config> Pallet<T> {
         match T::EthereumPublicKeyChecker::get_validator_for_eth_public_key(
             &recovered_public_key.expect("Checked for error"),
         ) {
-            Some(maybe_validator) => {
-                maybe_validator == validator.account_id},
+            Some(maybe_validator) => maybe_validator == validator.account_id,
             _ => false,
         }
     }
--- a/pallets/validators-manager/src/mock.rs
+++ b/pallets/validators-manager/src/mock.rs
@@ -467,8 +467,6 @@ thread_local! {
 
     static MOCK_TX_ID: RefCell<TransactionId> = RefCell::new(INITIAL_TRANSACTION_ID);
 
-    pub static ETH_PUBLIC_KEY_VALID: RefCell<bool> = RefCell::new(true);
-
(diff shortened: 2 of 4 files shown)


pallets/validators-manager/src/mock.rs
```diff
@@ -364,36 +356,12 @@ parameter_types! {
 
 impl CandidateTransactionSubmitter<AccountId> for TestRuntime {
     fn submit_candidate_transaction_to_tier1(
```
alexpArtos, line 358: What happened here? Is this function no longer needed then? Is it a placeholder now?

pallets/validators-manager/src/tests_voting_deregistration.rs
```diff
@@ -78,10 +87,10 @@ fn setup_ext_builder() -> (TestExternalities, Arc<RwLock<PoolState>>, Arc<RwLock
 
     (ext, pool_state, offchain_state)
 }
-
+/// Setups a voting session to deregister collator 5. Sender is collator 3
 fn setup_voting_session(action_id: &ActionId<AccountId>) {
     let collator_eth_public_key = ecdsa::Public::from_raw(hex!(
-        "02407b0d9f41148bbe3b6c7d4a62585ae66cc32a707441197fa5453abfebd31d57"
+        "0362c0a046dacce86ddd0343c6d3c7c79c2208ba0d9c9cf24a6d046d21d21f90f7"
```
alexpArtos, line 93: Can we turn these keys into constants? Perhaps it would make the code easier to follow, as in we'd know immediately which collator we're dealing with

pallets/validators-manager/src/tests_voting_deregistration.rs
```diff
@@ -425,7 +450,7 @@ mod approve_vote {
                 mock_response_of_get_ecdsa_signature(
                     &mut context.offchain_state.write(),
                     eth_compatible_data,
-                    Some(hex::encode([1; 65].to_vec()).as_bytes().to_vec()),
+                    Some(hex::encode([2; 65].to_vec()).as_bytes().to_vec()),
```
alexpArtos, line 453: Is this related to the index in `get_validator_by_index`?

alexpArtos (approved):

pallets/avn/src/lib.rs
```diff
@@ -252,7 +252,8 @@ impl<T: Config> Pallet<T> {
         match T::EthereumPublicKeyChecker::get_validator_for_eth_public_key(
             &recovered_public_key.expect("Checked for error"),
         ) {
-            Some(maybe_validator) => maybe_validator == validator.account_id,
+            Some(maybe_validator) => {
+                maybe_validator == validator.account_id},
```
ivan-cholakov, line 256: `cargo fmt`?

ivan-cholakov (approved):

pallets/avn/src/lib.rs
thadouk: this is with `cargo +nightly fmt` applied.

PR #169, opened 2023-06-16
Feature:  Update logic on adding and removing a collator

thadouk:
Updates the logic and tests of the validators manager pallet, to align with the smart contract interface.
Updates the way the confirmations are generated, reactors tests and benchmarks around that area.

Related JIRA tasks:
- SYS-3387
- SYS-3333
- SYS-3360


ivan-cholakov (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -33,9 +41,8 @@ impl EthTransactionType {
     pub fn to_abi(&self) -> Result<EthTransactionDescription, ethabi::Error> {
         match self {
             EthTransactionType::PublishRoot(d) => Ok(d.to_abi()),
-            EthTransactionType::DeregisterValidator(d) => Ok(d.to_abi()),
-            EthTransactionType::SlashValidator(d) => Ok(d.to_abi()),
-            EthTransactionType::ActivateValidator(d) => Ok(d.to_abi()),
+            EthTransactionType::ActivateCollator(d) => Ok(d.to_abi()),
+            EthTransactionType::DeregisterCollator(d) => Ok(d.to_abi()),
```
nahuseyoum, line 45: Is "re-enabling slashing" captured anywhere in JIRA? We seem to be removing that functionality here.

pallets/validators-manager/src/confirmations.rs
```diff
@@ -0,0 +1,60 @@
+use frame_support::log;
+use pallet_ethereum_transactions::ethereum_transaction::{
+    ActivateCollatorData, DeregisterCollatorData,
+};
+use sp_io::hashing::keccak_256;
+
+const PACKED_KEYS_SIZE: usize = 96;
+
+#[cfg(test)]
```
nahuseyoum, line 9: Do we want this here?

pallets/validators-manager/src/tests_voting_deregistration.rs
```diff
@@ -93,24 +106,36 @@ fn setup_voting_session(action_id: &ActionId<AccountId>) {
     assert_eq!(ValidatorManager::get_vote(action_id).nays.is_empty(), true);
 }
 
+// TODO use the private keys of the authorities to sign the _eth_compatible_data
+fn create_valid_signed_respose(
```
nahuseyoum, line 110:
```suggestion
fn create_valid_signed_response(
```

nahuseyoum (approved):
LGTM but we should make a decision on the Slashing logic. I think we are implicitly removing it and I am not sure if that is a good idea.

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
```diff
@@ -2,16 +2,21 @@
 
 use crate::ethereum_transaction::*;
 use ethabi::{Function, Param, ParamType, Token};
+use sp_core::H512;
 
 pub const ROOT_HASH: [u8; 32] = [3; 32];
 pub const T2_PUBLIC_KEY: [u8; 32] = [4; 32];
+pub const T1_PUBLIC_KEY: [u8; 64] = [5u8; 64];
```
fluorostani, line 9: we should be consistent over whether we specify the type or not here

fluorostani (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
thadouk: yes we have a ticket for it.

pallets/validators-manager/src/confirmations.rs
nahuseyoum: Move it to lib.rs with the rest of the tests

thadouk: Why? The context is much better this way. This file has this set of tests for these functions.

pallets/ethereum-transactions/src/tests/tests_eth_transaction_type.rs
thadouk: I think I get it now, you mean `[5u8; 64]` vs `[5; 64]`
I prefer the first way, it's more explicit.

pallets/validators-manager/src/confirmations.rs
nahuseyoum: Having these paths scattered all over the place will make it harder to see where they are. What is the problem of having them in 1 place, grouped together? What is the rational to change this existing practise?

thadouk: The rationale behind this proposal is to improve modularity within our codebase. Currently, the lib.rs file is responsible for organizing modules and tests. Instead, a simpler approach would be for each module to specify the path to its test files as a self-contained unit. It is a more modular approach I think.

PR #170, opened 2023-06-21
Fixes benchmarks for validators manager pallet 

thadouk:
This commit addresses the issues with the benchmarks in the validators manager pallet.
The benchmarks have been updated to improve their setup. Now, the tests will add
additional collators on top of the existing genesis configuration.
It is required to have at least two additional collators: one collator for the
resignation action and another collator for casting a vote for the action (vote_sender)
and sending the transaction to Ethereum.

The remaining collators generated randomly using a seed are not used in any action.
Additionally, this commit refactors and simplifies the interface of the benchmark tests
to enhance their readability and maintainability.

Lastly, the project's dependencies are updated to be used when needed.

Related JIRA tickets:
- SYS-3387
- SYS-3333
- SYS-3360

### DIFF
--- a/pallets/validators-manager/src/benchmarking.rs
+++ b/pallets/validators-manager/src/benchmarking.rs
@@ -16,15 +16,15 @@ use pallet_parachain_staking::{Currency, Pallet as ParachainStaking};
 use pallet_session::Pallet as Session;
 use secp256k1::{PublicKey, SecretKey};
 use sp_avn_common::eth_key_actions::decompress_eth_public_key;
-use sp_core::{ecdsa::Public, sr25519, Pair, H512};
+use sp_core::{ecdsa::Public, H512};
 
-const RESIGNING_COLLATOR_PRIVATE_KEY: [u8; 32] = [6u8; 32];
+// Resigner keys derived from [6u8; 32] private key
 const RESIGNING_COLLATOR_PUBLIC_KEY_BYTES: [u8; 32] =
     hex!["ea3021db7da7831e0d5ed7e60a8102d2d721bcca88adb03ee992f4dec3baee3e"];
 const RESIGNING_COLLATOR_ETHEREUM_PUBLIC_KEY: [u8; 33] =
     hex!["03f006a18d5653c4edf5391ff23a61f03ff83d237e880ee61187fa9f379a028e0a"];
 
-const VOTING_COLLATOR_PRIVATE_KEY: [u8; 32] = [7u8; 32];
+// Vote sender keys derived from [7u8; 32] private key
 const VOTING_COLLATOR_PUBLIC_KEY_BYTES: [u8; 32] =
     hex!["7c0f469d3bd340bae718203fa30ca071a5e37c751e891dbded837b213d45d91d"];
 const VOTING_COLLATOR_ETHEREUM_PUBLIC_KEY: [u8; 33] =
@@ -40,10 +40,8 @@ fn generate_resigning_collator_account_details<T: Config>(
     let authority_id =
         <T as avn::Config>::AuthorityId::generate_pair(Some("//avn_resigner".as_bytes().to_vec()));
     let eth_public_key = Public::from_raw(RESIGNING_COLLATOR_ETHEREUM_PUBLIC_KEY);
-    let account_bytes = sr25519::Pair::from_seed(&RESIGNING_COLLATOR_PRIVATE_KEY).public().to_vec();
-    assert_eq!(account_bytes.as_slice(), RESIGNING_COLLATOR_PUBLIC_KEY_BYTES.as_slice());
-
-    let account_id = T::AccountId::decode(&mut account_bytes.as_slice()).unwrap();
+    let account_id =
+        T::AccountId::decode(&mut RESIGNING_COLLATOR_PUBLIC_KEY_BYTES.as_slice()).unwrap();
 
     (account_id, authority_id, eth_public_key)
 }
@@ -53,10 +51,8 @@ fn generate_sender_collator_account_details<T: Config>(
     let authority_id =
         <T as avn::Config>::AuthorityId::generate_pair(Some("//avn_sender".as_bytes().to_vec()));
     let eth_public_key = Public::from_raw(VOTING_COLLATOR_ETHEREUM_PUBLIC_KEY);
-    let account_bytes = sr25519::Pair::from_seed(&VOTING_COLLATOR_PRIVATE_KEY).public().to_vec();
-    assert_eq!(account_bytes.as_slice(), VOTING_COLLATOR_PUBLIC_KEY_BYTES.as_slice());
-
-    let account_id = T::AccountId::decode(&mut account_bytes.as_slice()).unwrap();
+    let account_id =
+        T::AccountId::decode(&mut VOTING_COLLATOR_PUBLIC_KEY_BYTES.as_slice()).unwrap();
 
     (account_id, authority_id, eth_public_key)
 }
@@ -77,9 +73,7 @@ fn setup_additional_validators<T: Config>(number_of_additional_validators: u32) 
             i if i == vote_sender_index => generate_sender_collator_account_details::<T>(),
             _ => (
                 account("dummy_validator", i, i),
-                <T as avn::Config>::AuthorityId::generate_pair(Some(
-                    format!("//avn_key_{:?}", i).as_bytes().to_vec(),
-                )),
+                <T as avn::Config>::AuthorityId::generate_pair(None),
                 generate_collator_eth_public_key_from_seed::<T>(i as u64),
             ),
         };


pallets/validators-manager/Cargo.toml
```diff
@@ -62,11 +63,15 @@ std = [
...
+	"pallet-parachain-staking/std",
+	"libsecp256k1?/std",
 ]
 runtime-benchmarks = [
     "frame-benchmarking/runtime-benchmarks",
     "frame-support/runtime-benchmarks",
     "frame-system/runtime-benchmarks",
-    "pallet-ethereum-transactions/runtime-benchmarks"
+    "pallet-ethereum-transactions/runtime-benchmarks",
+	"rand",
```
ivan-cholakov, line 74: `cargo fmt`?

thadouk: I don't think fmt picks up formating within cargo toml files

ivan-cholakov (approved):

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
+    hex!["02989c0b76cb563971fdc9bef31ec06c3560f3249d6ee9e5d83c57625596e05f6f"];
+
+const MINIMUM_ADDITIONAL_BENCHMARKS_VALIDATORS: usize = 2;
+
+const MNEMONIC: &str =
+    "basic anxiety marine match castle rival moral whisper insane away avoid bike";
+
+fn generate_resigning_collator_account_details<T: Config>(
+) -> (T::AccountId, <T as pallet_avn::Config>::AuthorityId, Public) {
+    let key = <T as avn::Config>::AuthorityId::generate_pair(Some(MNEMONIC.as_bytes().to_vec()));
```
alexpArtos, line 39: Why are we using the same mnemonic for the Resigner and the Sender? I think we should rename the constant to make clear what chain and user it belongs to.

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
-    avn::Validators::<T>::put(validators.clone());
+fn generate_sender_collator_account_details<T: Config>(
+) -> (T::AccountId, <T as pallet_avn::Config>::AuthorityId, Public) {
+    let key = <T as avn::Config>::AuthorityId::generate_pair(Some(MNEMONIC.as_bytes().to_vec()));
 
-    validators
-        .iter()
-        .enumerate()
-        .for_each(|(i, v)| force_add_collator::<T>(&v.account_id, i as u64));
+    let mut resigner_account_bytes: [u8; 32] = [0u8; 32];
```
alexpArtos, line 55:
```suggestion
    let mut sender_account_bytes: [u8; 32] = [0u8; 32];
```

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
+    resigner_account_bytes
```
alexpArtos, line 56:
```suggestion
    sender_account_bytes
```

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
+        .copy_from_slice(&hex!("b41f90b123b66c18f0f869f3b9ae8a09d118419f8736240fcc7b8256517cc233"));
```
alexpArtos, line 57: Ditto

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
 
-    // Setup resigner account id and key
-    let resigner: Validator<T::AuthorityId, T::AccountId> = validators[1].clone();
     let mut resigner_account_bytes: [u8; 32] = [0u8; 32];
     resigner_account_bytes
-        .copy_from_slice(&hex!("1ed1aadead9704b693af012a9f24e1f00dc7e2a0b4eb99f9e0bc0c35a8d20223"));
-    let resigner_account_id =
-        T::AccountId::decode(&mut &resigner_account_bytes.encode()[..]).unwrap();
-    validators[1] = Validator::new(resigner_account_id, resigner.key);
+        .copy_from_slice(&hex!("0e08c08252ef2a5cbeb5d6a0dcaa1b8354aded57d603cdc4adc5e27709a6904f"));
```
alexpArtos, line 43: Can we make this a constant? How does this relate to the key in line 47?
Also, could we make clear in the variable name (`resigner_account_bytes`) what chain this account is in? And is this basically the account **public** key?

pallets/validators-manager/src/benchmarking.rs
alexpArtos: Same comments apply as for the resigner function.

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -290,23 +324,24 @@ benchmarks! {
...
 
     remove_validator {
-        let v in (DEFAULT_MINIMUM_VALIDATORS_COUNT as u32 + 1) .. MAX_VALIDATOR_ACCOUNT_IDS;
+        let v in (MINIMUM_ADDITIONAL_BENCHMARKS_VALIDATORS as u32 + 1) .. MAX_VALIDATOR_ACCOUNT_IDS;
 
-        let validators = setup_validators::<T>(v);
-        let caller = validators[(v - 1) as usize].account_id.clone();
-    }: remove_validator(RawOrigin::Root, caller.clone())
+        setup_additional_validators::<T>(v);
+        // let caller = validators[(v - 1) as usize].account_id.clone();
```
nahuseyoum, line 330:
```suggestion
```

nahuseyoum (approved):

pallets/validators-manager/Cargo.toml
```diff
@@ -13,7 +13,7 @@ serde = { version = "1.0.137", optional = true }
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
 hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
 scale-info = { version = "2.0", default-features = false, features = ["derive"] }
-rand = { version = "0.7.2", default-features = false }
+rand = { version = "0.7.2", default-features = false, optional = true }
```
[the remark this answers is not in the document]
thadouk: conditionally runtime, in benchmarks, hence optiona.

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
+    assert!(number_of_additional_validators >= MINIMUM_ADDITIONAL_BENCHMARKS_VALIDATORS as u32);
+
+    let mut avn_validators: Vec<Validator<<T as pallet_avn::Config>::AuthorityId, T::AccountId>> =
+        Vec::new();
+
+    let mut validators: Vec<(T::AccountId, Public)> = Vec::new();
+    let vote_sender_index = number_of_additional_validators - (1 as u32);
+
+    for i in 0..number_of_additional_validators {
+        let (account, key, eth_key) = match i {
```
alexpArtos, line 76: Could we make clear whether `key` is a private or public key and on which chain it is?

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
+            0 => generate_resigning_collator_account_details::<T>(),
+            i if i == vote_sender_index => generate_sender_collator_account_details::<T>(),
+            _ => (
+                account("dummy_validator", i, i),
+                <T as avn::Config>::AuthorityId::generate_pair(Some(MNEMONIC.as_bytes().to_vec())),
```
alexpArtos, line 81: are we always using the same private key for all collators?

pallets/validators-manager/Cargo.toml
```diff
@@ -29,7 +29,8 @@ sp-io = { default-features = false, git = "https://github.com/paritytech/substra
 sp-application-crypto = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36" }
 sp-staking = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36" }
 hex-literal = { version = "0.3.4", default-features = false }
-
+libsecp256k1 = { version = "0.3.2", default-features = false, features = ["hmac"], optional = true }
```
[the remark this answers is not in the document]
thadouk: The `decompress_eth_public_key` function is part of `sp-avn-common` so technically is not a direct dependency of the pallet. It is a dependency though when using benchmarks.

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -10,102 +10,129 @@ use super::*;
...
-        action_id.action_account_id.clone(),
-        action_id.ingress_counter,
-    );
+    let signature: <T::AuthorityId as RuntimeAppPublic>::Signature = generate_signature::<T>();
+    let quorum = setup_voting_session::<T>(&action_id);
+    // signed by sender private key [7u8; 32]
+    let approval_signature: ecdsa::Signature = ecdsa::Signature::from_slice(&hex!("c951051534f26f2101224ed0ab56ddaeffbb57952e06166c5033fa2290d8306f046628ede2a45cff90391d0a604f08989ced0177c63febbe3afd5340c38561091c")).unwrap().into();
 
-    (sender, action_id, approval_signature, signature, quorum)
+    setup_resignation_action_data::<T>(vote_sender_id.clone(), action_id.ingress_counter);
```
alexpArtos, line 120: I'm not sure about this. Could we have the test be clearer and make sure our intentions are correct?

If I read correctly, we have a variable action_id as defined by the `generate_xxx_account_details` function(s). But we create a signature for a fixed value of `0`. We also have an ingress counter of `1`. So, it's a mystery to me how these values are related, and I think some renaming can explain that.

So, what is this signature confirming? Why is it fixed at 0? Does it depend on the value of action_id we have in here?

I think it would be good for the `signature` variable to reflect its intentions, for example, `signature_for_current_action_id`.

alexpArtos: And why/how do we know the signer is private key [7u8; 32]? From what I see, it should be `T::AuthorityId::generate_pair(None);`. Is this fixed? Why is this account signing and where do we verify the signature? How do we ensure it is verified by the same account?

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -411,8 +446,8 @@ benchmarks! {
         let o in 1 .. MAX_OFFENDERS; // maximum num of offenders need to be less than one third of minimum validators so the benchmark won't panic
 
         let number_of_validators = MAX_VALIDATOR_ACCOUNT_IDS;
-        let validators = setup_validators::<T>(number_of_validators);
-        let (sender, action_id, _, signature, quorum) = setup_action_voting::<T>(validators.clone());
+        let validators = setup_additional_validators::<T>(number_of_validators);
```
alexpArtos, line 448: Apparently, this function no longer returns any value.

pallets/validators-manager/src/benchmarking.rs
```diff
@@ -446,8 +481,8 @@ benchmarks! {
         let o in 1 .. MAX_OFFENDERS; // maximum of offenders need to be less one third of minimum validators so the benchmark won't panic
 
         let number_of_validators = MAX_VALIDATOR_ACCOUNT_IDS;
-        let validators = setup_validators::<T>(number_of_validators);
-        let (sender, action_id, _, signature, quorum) = setup_action_voting::<T>(validators.clone());
+        let validators = setup_additional_validators::<T>(number_of_validators);
```
alexpArtos, line 483: ditto

pallets/validators-manager/src/benchmarking.rs
thadouk: updated, we use accounts generated from "//<name>" now for all generated accounts

pallets/validators-manager/src/benchmarking.rs
thadouk: Updated

pallets/validators-manager/src/benchmarking.rs
thadouk: I think this goes beyond fixing the benchmarks and probably a general refactor to make this area easier to the eye.
An action though is defined by a validator and an ingress, that action will need some data, that will be identified from that key.
The action can be collator activation or deregistration, in this case, we setup a deregistration.
As for the `signature` variable, this is the unsigned extrinsic signature we attach, probably we don't need to validate it during benchmarks, hence the "dummy" values with 0.

alexpArtos (approved):

PR #174, opened 2023-07-06
SYS-3453 Eliminated token-manager transfer functionality to remove duplicated transfer events

Xakep1101:
Removed transfer event functionality from token-manager pallet as the event already exists (balances.Transfer)

[SYS-3453]()

[SYS-3453]: https://aventus-network-services.atlassian.net/browse/SYS-3453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nahuseyoum (commented):
Is this what we want to do here? I thought all we wanted was to remove the extra event not the AVT flow from token manager

alexpArtos (changes requested):
Hi Michael, I believe this was not the spirit of the task. This PR seems to be removing a whole and very useful operation, whereas all we want is to remove a duplicated event.
We should sync offline.

PR #175, opened 2023-07-06
SYS-3338: Convert ValidatorAccountIds to Bounded Vector

thadouk:
This commit converts the ValidatorAccountIds storage item to a bounded Vector.
The vector is now limited to 256 elements. The necessary adjustments have been
made to the operations that interact with the storage, using the appropriate interfaces.
No storage migration is required for this conversion.
Tests have been updated accordingly to reflect the changes.

### DIFF
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -47,7 +47,7 @@ use sp_avn_common::{
     calculate_two_third_quorum, eth_key_actions::decompress_eth_public_key, event_types::Validator,
     safe_add_block_numbers, IngressCounter,
 };
-use sp_core::{ecdsa, H512};
+use sp_core::{ecdsa, H512, ConstU32};
 
 pub use pallet_parachain_staking::{self as parachain_staking, BalanceOf, PositiveImbalanceOf};
 use pallet_session::historical::IdentificationTuple;
@@ -174,7 +174,7 @@ pub mod pallet {
     #[pallet::storage]
     #[pallet::getter(fn validator_account_ids)]
     pub type ValidatorAccountIds<T: Config> =
-        StorageValue<_, BoundedVec<T::AccountId, ConstU32<MAXIMUM_VALIDATORS_BOUND>>>;
+        StorageValue<_, BoundedVec<T::AccountId, MaximumValidatorsBound>>;
 
     #[pallet::storage]
     pub type ValidatorActions<T: Config> = StorageDoubleMap<
@@ -588,7 +588,7 @@ const MAX_OFFENDERS: u32 = 2;
 // TODO [TYPE: review][PRI: medium]: if needed, make this the default value to a configurable
 // option. See MinimumValidatorCount in Staking pallet as a reference
 const DEFAULT_MINIMUM_VALIDATORS_COUNT: usize = 2;
-const MAXIMUM_VALIDATORS_BOUND: u32 = 256;
+type MaximumValidatorsBound = ConstU32<256>;
 
 const NAME: &'static [u8; 17] = b"validatorsManager";
 
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -257,6 +257,7 @@ pub mod pallet {
             ensure_root(origin)?;
             let validator_account_ids =
                 Self::validator_account_ids().ok_or(Error::<T>::NoValidators)?;
+            ensure!(validator_account_ids.len() > 0, Error::<T>::NoValidators);
 
             ensure!(
                 !validator_account_ids.contains(&collator_account_id),
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -47,7 +47,7 @@ use sp_avn_common::{
     calculate_two_third_quorum, eth_key_actions::decompress_eth_public_key, event_types::Validator,
     safe_add_block_numbers, IngressCounter,
 };
-use sp_core::{ecdsa, H512, ConstU32};
+use sp_core::{ecdsa, ConstU32, H512};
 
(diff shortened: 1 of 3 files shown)


pallets/validators-manager/src/lib.rs
```diff
@@ -192,6 +190,7 @@ pub mod pallet {
 
     #[pallet::storage]
     #[pallet::getter(fn get_vote)]
+    #[pallet::unbounded]
```
alexpArtos, line 193: why do we need this?

pallets/validators-manager/src/lib.rs
```diff
@@ -257,8 +256,7 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             ensure_root(origin)?;
             let validator_account_ids =
-                Self::validator_account_ids().or_else(|| Some(vec![])).expect("empty vec");
-            ensure!(validator_account_ids.len() > 0, Error::<T>::NoValidators);
+                Self::validator_account_ids().ok_or(Error::<T>::NoValidators)?;
```
alexpArtos, line 259: why did we remove the explicit check that the length is > 0? What exactly does `ok_or` do in this case?

pallets/validators-manager/src/tests/tests.rs
```diff
@@ -588,7 +588,9 @@ mod add_validator {
                 let context = &AddValidatorContext::default();
 
                 set_session_keys(&context.collator);
-                <<ValidatorManager as Store>::ValidatorAccountIds>::append(&context.collator);
+                assert_ok!(<<ValidatorManager as Store>::ValidatorAccountIds>::try_append(
```
alexpArtos, line 591: Shouldn't we add a test in here for when we have the maximum number of collators and try to add beyond it?

thadouk: What is the behaviour the test is testing?
The BoundedVector implementation?

pallets/validators-manager/src/lib.rs
thadouk:
```
/// The optional attribute `#[pallet::unbounded]` declares the storage as unbounded. When
/// implementating the storage info (when `#[pallet::generate_storage_info]` is specified on
/// the pallet struct placeholder), the size of the storage will be declared as unbounded. This
/// can be useful for storage which can never go into PoV (Proof of Validity).
```

ivan-cholakov (approved):

pallets/validators-manager/src/tests/tests.rs
alexpArtos: No, actually not. We are doing functional testing, or in other words, behaviour-testing.

We are testing `add_validator`. If you see the sequence of contexts in the code, it would be something like this:

`add_validator fails_when validator_list_is_full`

So, you see, from the functionality's point of view, we don't care if we have bounded-vec or not. We now have a functional difference to what we had before, (a limit on addition) but we don't have that in the tests. In this case, the correct behaviour would be to receive a MaxValidators-type error.

It's a different function contract, so I think we should have a test to represent that.

thadouk: I see your point, although I think the test is of little value. We don't do such extensive testing on our endpoints.
The mock effort required trumps the value the test adds, as reaching this number of collator is academic.
The bound is not there to add such functionality to the system, but to align the implementation to the framework guidelines.
That being said I will add the test to accommodate the request as it probably won't be that complicated.
This won't be the case though for every bounded vector we convert.

pallets/validators-manager/src/lib.rs
```diff
@@ -234,7 +233,7 @@ pub mod pallet {
                 self.validators.len()
             );
             for (validator_account_id, eth_public_key) in &self.validators {
-                <ValidatorAccountIds<T>>::append(validator_account_id);
+                assert_ok!(<ValidatorAccountIds<T>>::try_append(validator_account_id));
```
nahuseyoum, line 236:
```suggestion
                <ValidatorAccountIds<T>>::try_append(validator_account_id).expect("Too many validator accounts in genesis");
```

pallets/validators-manager/src/lib.rs
```diff
@@ -66,13 +66,13 @@ use crate::offence::{
 #[frame_support::pallet]
 pub mod pallet {
     use super::*;
-    use frame_support::pallet_prelude::*;
+    use frame_support::{assert_ok, pallet_prelude::*};
```
nahuseyoum, line 69:
```suggestion
    use frame_support::pallet_prelude::*;
```

alexpArtos (approved):

pallets/validators-manager/src/lib.rs
alexpArtos: I think applies in a Map, not a Vec. We are not making maps bounded.

nahuseyoum: If all we have is a map then we shouldn't use this attribute because the idea behind it is to explicitly mention that a storage has `unbounded` items and is usually used with storage items that store vectors.

thadouk: The storage item is unbounded cause the map value items, contain unbounded vectors. This is addressed in the voting changes in #178 
The attribute is optional, and it was added to be informative.
Nonetheless, I can see that since it creates confusion, I will remove it since it is removed anyway on #178

pallets/validators-manager/src/lib.rs
thadouk: I prefer the assertion as it is more compact.

nahuseyoum: But the assertion doesn't give you a reason for the error. This one is more explicit and provides a meaningful error.

nahuseyoum (approved):

PR #176, opened 2023-07-07
SYS-3453: Remove duplicate transfer event and update tests

Xakep1101:
Removed the 'TokenTransferred' event from the  token-manager pallet and updated tests accordingly

### DIFF
--- a/pallets/token-manager/src/lib.rs
+++ b/pallets/token-manager/src/lib.rs
@@ -161,12 +161,6 @@ pub mod pallet {
             token_balance: T::TokenBalance,
             eth_tx_hash: H256,
         },
-        TokenTransferred {
-            token_id: T::TokenId,
-            sender: T::AccountId,
-            recipient: T::AccountId,
-            token_balance: T::TokenBalance,
-        },
         CallDispatched {
             relayer: T::AccountId,
             call_hash: T::Hash,
@@ -315,12 +309,6 @@ pub mod pallet {
 
             Self::settle_transfer(&token_id, &from, &to, &amount)?;
 
-            Self::deposit_event(Event::<T>::TokenTransferred {
-                token_id,
-                sender: from,
-                recipient: to,
-                token_balance: amount,
-            });
             Ok(())
         }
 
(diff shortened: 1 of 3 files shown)


pallets/token-manager/src/lib.rs
```diff
@@ -161,12 +161,6 @@ pub mod pallet {
             token_balance: T::TokenBalance,
             eth_tx_hash: H256,
         },
-        TokenTransferred {
-            token_id: T::TokenId,
-            sender: T::AccountId,
-            recipient: T::AccountId,
-            token_balance: T::TokenBalance,
-        },
```
alexpArtos, line 169: I don't think you want to remove this.

pallets/token-manager/src/test_non_avt_tokens.rs
```diff
@@ -234,14 +234,6 @@ fn avn_test_signed_transfer_with_valid_input_should_succeed() {
...
             4 * amount
         );
-
-        assert!(System::events().iter().any(|a| a.event ==
-            RuntimeEvent::TokenManager(crate::Event::<TestRuntime>::TokenTransferred {
-                token_id: NON_AVT_TOKEN_ID,
-                sender: sender_account_id,
-                recipient: recipient_account_id,
-                token_balance: amount
-            })));
```
alexpArtos, line 244: We should be changing only the logic for AVT transfers. That means the tests for NON-AVT tokens should not be touched.

pallets/token-manager/src/test_proxying_signed_transfer.rs
```diff
@@ -175,19 +175,12 @@ fn check_proxy_transfer_default_call_succeed(call: Box<<TestRuntime as Config>::
...
             relayer: default_relayer(),
             call_hash
         })));
-    assert!(System::events().iter().any(|a| a.event ==
-        RuntimeEvent::TokenManager(crate::Event::<TestRuntime>::TokenTransferred {
-            token_id: NON_AVT_TOKEN_ID,
-            sender: default_sender(),
-            recipient: default_receiver(),
-            token_balance: DEFAULT_AMOUNT
-        })));
```
alexpArtos, line 190: Ditto

alexpArtos (changes requested):
I think this still needs another iteration. Let's pair on it.

PR #177, opened 2023-07-07
SYS-3453: Removed duplicate transfer event and updated tests

Xakep1101:
Removed duplicate transfer events by refactoring code in lib.rs. Updated tests to test for balances.Transfer instead of token-manager.TokenTransferred when testing AVT transfers.


alexpArtos (approved):
Great work, Michael. Looks much cleaner now. Congrats for your first code and test changes.
I trust you have run the tests, and they all pass?

nahuseyoum (approved):
Since we are changing this, shall we make the same changes to Lower? i.e. only raise TokenLowered when its not AVT?

PR #178, opened 2023-07-11
SYS-3338: Convert Voting to Use Bounded Data Structures

thadouk:
This commit converts the Voting mechanism in the summary and validators manager to use Bounded data structures. The changes include:

- Using BoundedVec for votes and confirmations
- Introducing WeakBoundedVec for the voting session ID (to be updated to BoundedVec in a separate change)
- Enabling storage information generation for the summary and validators manager pallets
- Adjusting function interfaces to handle potential failure when adding a vote
- Replacing Infallible paths with WeakBoundedVec as an interim solution
- Updating tests and benchmarks
- Resolving a dependency issue in benchmarks

### DIFF
--- a/pallets/avn/src/vote.rs
+++ b/pallets/avn/src/vote.rs
@@ -2,16 +2,17 @@
 
 #[cfg(not(feature = "std"))]
 extern crate alloc;
+
 #[cfg(not(feature = "std"))]
 use alloc::string::{String, ToString};
 
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::{
     dispatch::{DispatchError, DispatchResult},
     ensure,
 };
 use sp_application_crypto::RuntimeAppPublic;
-use sp_avn_common::event_types::Validator;
+use sp_avn_common::{event_types::Validator, MaximumValidatorsBound, VotingSessionIdBound};
 use sp_core::ecdsa;
 use sp_runtime::{
     scale_info::TypeInfo,
@@ -19,6 +20,7 @@ use sp_runtime::{
     transaction_validity::{
         InvalidTransaction, TransactionPriority, TransactionValidity, ValidTransaction,
     },
+    BoundedVec, WeakBoundedVec,
 };
 use sp_std::prelude::*;
 
@@ -32,20 +34,21 @@ pub const VOTING_SESSION_DATA_IS_NOT_FOUND: u8 = 5;
 pub const APPROVE_VOTE: bool = true;
 pub const REJECT_VOTE: bool = false;
 
-#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo)]
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
 pub struct VotingSessionData<AccountId, BlockNumber> {
+    // TODO convert to BoundedVec
     /// The unique identifier for this voting session
-    pub voting_session_id: Vec<u8>,
+    pub voting_session_id: WeakBoundedVec<u8, VotingSessionIdBound>,
     /// The number of approval votes that are needed to reach an outcome.
     pub threshold: u32,
     /// The current set of voters that approved it.
-    pub ayes: Vec<AccountId>,
+    pub ayes: BoundedVec<AccountId, MaximumValidatorsBound>,
     /// The current set of voters that rejected it.
-    pub nays: Vec<AccountId>,
+    pub nays: BoundedVec<AccountId, MaximumValidatorsBound>,
     /// The hard end time of this vote.
     pub end_of_voting_period: BlockNumber,
     /// The confirmations collected from the aye votes
-    pub confirmations: Vec<ecdsa::Signature>,
+    pub confirmations: BoundedVec<ecdsa::Signature, MaximumValidatorsBound>,
     /// The block number this session was created on
     pub created_at_block: BlockNumber,
 }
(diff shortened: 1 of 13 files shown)


pallets/avn/src/vote.rs
```diff
@@ -32,20 +34,21 @@ pub const VOTING_SESSION_DATA_IS_NOT_FOUND: u8 = 5;
 pub const APPROVE_VOTE: bool = true;
 pub const REJECT_VOTE: bool = false;
 
-#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo)]
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
 pub struct VotingSessionData<AccountId, BlockNumber> {
+    // TODO convert to BoundedVec
     /// The unique identifier for this voting session
-    pub voting_session_id: Vec<u8>,
+    pub voting_session_id: WeakBoundedVec<u8, VotingSessionIdBound>,
```
ivan-cholakov, line 41: why do we have to use `WeakBoundedVec`

ivan-cholakov (approved):

pallets/avn/src/vote.rs
nahuseyoum, line 41: Why not convert to boundedVec now?

pallets/summary/src/vote.rs
```diff
@@ -95,15 +95,33 @@ impl<T: Config> VotingSessionManager<T::AccountId, T::BlockNumber> for RootVotin
...
-        })
+    fn record_approve_vote(
+        &self,
+        voter: T::AccountId,
+        approval_signature: ecdsa::Signature,
+    ) -> DispatchResult {
+        <Summary<T> as Store>::VotesRepository::try_mutate(
+            &self.root_id,
+            |vote| -> DispatchResult {
+                vote.ayes.try_push(voter).map_err(|_| avn_error::<T>::VectorBoundsExceeded)?;
```
nahuseyoum, line 106: :+1:

primitives/avn-common/src/lib.rs
```diff
@@ -38,6 +38,10 @@ pub const ETHEREUM_PREFIX: &'static [u8] = b"\x19Ethereum Signed Message:\n32";
 pub const EXTERNAL_SERVICE_PORT_NUMBER_KEY: &'static [u8; 15] = b"avn_port_number";
 /// Default port number the external service runs on.
 pub const DEFAULT_EXTERNAL_SERVICE_PORT_NUMBER: &str = "2020";
+/// Bound used for Vectors containing validators
+pub type MaximumValidatorsBound = ConstU32<256>;
```
nahuseyoum, line 42: If you add these types on the config of the pallets, you can query them from polkadotJs and via a script. The way these are set, we would have to see the code to know what the limit is which makes it harder for DApps to validate data.

pallets/avn/src/vote.rs
nahuseyoum: Won't this need a data migration? Because we are changing the inner type of a struct which will change how the data is encoded.

primitives/avn-common/src/lib.rs
thadouk: true, this will be the case in some of the pallets where there the bounded data input or output comes from a user.
So far, the operations refactored are either sudo or in the case of the voting, an internal mechanism that dapps and users don't use. It is a better fit to use const values in the code I think

pallets/avn/src/vote.rs
thadouk: > Why not convert to boundedVec now?

Next PR handles that.

> Won't this need a data migration? Because we are changing the inner type of a struct which will change how the data is encoded.

Good question. It is actually safe. The bounded vector implementation is the following:
```
/// A bounded vector.
///
/// It has implementations for efficient append and length decoding, as with a normal `Vec<_>`, once
/// put into storage as a raw value, map or double-map.
///
/// As the name suggests, the length of the queue is always bounded. All internal operations ensure
/// this bound is respected.
#[cfg_attr(feature = "std", derive(Serialize), serde(transparent))]
#[derive(Encode, scale_info::TypeInfo)]
#[scale_info(skip_type_params(S))]
pub struct BoundedVec<T, S>(
	pub(super) Vec<T>,
	#[cfg_attr(feature = "std", serde(skip_serializing))] PhantomData<S>,
);
```
BoundedVec is a tuple struct of Vector and PhantomData.
PhantomData is a marker type so it doesn't get included during the scale encoding of the data.

The `skip_type_params` attribute also skips the field from generating type information for the field as well.
https://docs.rs/scale-info/latest/scale_info/#scale_infoskip_type_params

The `serde(skip_serializing)` attribute excludes the field when performing a serde serialisation.

From my understanding the encoding excludes the const type that applies the bound, and the code implementation is responsible to use the correct bound otherwise the decoding will fail.

In sort the encoding is the same as having a vector, and there isn't a concern when you change it nor a data migration is needed.
We can put this to the test:
```
    #[test]
    fn encoding_tests() {
        let vector: Vec<u8> = vec![1,2,3];
        let bounded = BoundedVec::<u8,ConstU32<3>>::truncate_from(vector.clone());
        assert_eq!(vector.encode(), bounded.encode());
    }
```
I have also successfully upgrade the chain with summary data. Apart from a UI refresh the chain had no issue with the upgrade

primitives/avn-common/src/lib.rs
nahuseyoum: My suggestion was not related to the source of the bounded data but rather to making that accessible in a running chain. For example if you want to know the max number of validators the voting logic supports, you would have to find the code and check the value. My suggestion is to make that value accessible in the chain state so you can read it from polkadotJS or from a script programatically. This approach is DApp friendly, user friendly and incurs no extra cost. 
image

nahuseyoum: Discussed offline and agreed to make this change in a separate PR

nahuseyoum (approved):

PR #180, opened 2023-07-12
SYS-3338: Convert Voting Session ID to Use Bounded Data Structures

thadouk:
Updates the Voting module to use `BoundedVec` instead of `WeakBoundedVec` for voting session IDs. The session IDs are now truncated using the `truncate` function to enforce the bounds of the data structure. It is important to ensure that the encoding of session IDs does not exceed the specified bounds.
To ensure this, tests have been added to verify that the constraints are properly enforced.

### DIFF
--- a/pallets/validators-manager/src/lib.rs
+++ b/pallets/validators-manager/src/lib.rs
@@ -19,7 +19,7 @@ use sp_runtime::{
     scale_info::TypeInfo,
     traits::{Convert, Member},
     transaction_validity::{InvalidTransaction, TransactionSource, TransactionValidity},
-    DispatchError, WeakBoundedVec,
+    DispatchError,
 };
 use sp_std::prelude::*;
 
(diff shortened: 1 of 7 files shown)


ivan-cholakov (approved):

pallets/validators-manager/src/lib.rs
```diff
@@ -1052,8 +1052,8 @@ impl<AccountId: Member + Encode> ActionId<AccountId> {
     fn new(action_account_id: AccountId, ingress_counter: IngressCounter) -> Self {
         return ActionId::<AccountId> { action_account_id, ingress_counter }
     }
-    fn session_id(&self) -> WeakBoundedVec<u8, VotingSessionIdBound> {
-        WeakBoundedVec::force_from(self.encode(), Some("validators manager action id"))
+    fn session_id(&self) -> BoundedVec<u8, VotingSessionIdBound> {
+        BoundedVec::truncate_from(self.encode())
```
nahuseyoum, line 1056: Wouldn't it be better to hash the value so we guarantee the size of the output?

thadouk: Great question and the answer is yes.
The VotingSessionIdBound is set to 64 bytes to hint at a blake512 hash for future implementations of the voting mechanism.

We had an internal discussion about this, and we opted for this approach as it doesn't require a conversion of the existing data, hence it's less risky and smaller change.

I am happy though to do a change on top of this and introduce the migration of all existing data to use a hash and constrain the sessionId to use a default implementation that uses the blake512 hash.
If we were to do this, we could also switch to a 32 bytes bound and use blake256 hash for the data.

nahuseyoum (approved):
Discussed the hashing comment offline and agreed to try and implement it in a separate PR.

alexpArtos (approved):

PR #182, opened 2023-07-17
SYS-3338 - converted Vec params to BoundedVec params

ivan-cholakov:
Converted the returned signature params of transfer,lift,lower functions

### DIFF
--- a/pallets/token-manager/src/lib.rs
+++ b/pallets/token-manager/src/lib.rs
@@ -35,14 +35,14 @@ use sp_avn_common::{
     event_types::{AvtGrowthLiftedData, EthEvent, EventData, LiftedData, ProcessedEventHandler},
     verify_signature, CallDecoder, InnerCallValidator, Proof,
 };
-use sp_core::{H160, H256};
+use sp_core::{ConstU32, H160, H256};
 use sp_runtime::{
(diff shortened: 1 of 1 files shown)


pallets/token-manager/src/lib.rs
```diff
@@ -596,17 +597,19 @@ impl<T: Config> Pallet<T> {
...
-        return (
-            SIGNED_TRANSFER_CONTEXT,
-            proof.relayer.clone(),
-            from,
-            to,
-            token_id,
-            amount,
-            sender_nonce,
+    ) -> BoundedVec<u8, MaximumSignedParamsSize> {
+        return BoundedVec::truncate_from(
```
thadouk, line 601: Is this safe? Does the proof have a fixed size?

thadouk (commented):
I think the changes need to start from a storage item cause that is where the bounds should be applied.

pallets/token-manager/src/lib.rs
ivan-cholakov: the tests are passing with the `pub type MaximumSignedParamsSize = ConstU32<256>;` constant

ivan-cholakov: what other testing should be done here?

PR #188, opened 2023-07-24
SYS-3521 SYS-3522 Open source avn-service - part 3

nahuseyoum:


node/avn-service/src/web3_utils.rs
```diff
@@ -0,0 +1,192 @@
...
+/// indicate an uncompressed public key; this first byte is ignored when
+/// computing the hash.
+pub fn public_key_address(public_key: &PublicKey) -> Address {
+    let public_key = public_key.serialize_uncompressed();
+
+    debug_assert_eq!(public_key[0], 0x04);
+    let hash = keccak256(&public_key[1..]);
+
+    Address::from_slice(&hash[12..])
+}
```
thadouk, line 185: I think this area could improve if we replace the magic numbers with const variables.

thadouk (approved):

PR #191, opened 2023-07-26
SYS-3521 Update pallet and primitive references to be path based

nahuseyoum:


fluorostani (approved):

nahuseyoum:
> Looks good! One question about Cargo.lock, did we copy the original version during the migration or did we let cargo resolve the dependencies as pallets were added? If the latter, a lot of these dependencies will probably use a newer version that neither we nor parity have tested on this version of substrate/cumulus. If that is the case, I would prefer going with the first option, to have an as close as possible migration of the original node and its building dependencies.

The only changes in cargo.lock is the conversion of the avn pallets from packages to path references. No other library has been modified.

PR #193, opened 2023-07-29
SYS-3338 Adds migration for external reference

thadouk:
Convers UsedExternalReferences keys into WeakBoundedVector and adds migration that truncates and updates the keys where needed.

This is done as a stopgap solution, to transform any keys that exceed the bounds. Once all chains are successfully migrated, UsedExternalReferences will be reverted back to using Bounded Vectors

Review with conjunction to #194

### DIFF
--- a/pallets/nft-manager/src/lib.rs
+++ b/pallets/nft-manager/src/lib.rs
@@ -45,7 +45,7 @@ use sp_io::hashing::keccak_256;
 use sp_runtime::{
     scale_info::TypeInfo,
     traits::{Hash, IdentifyAccount, Member, Verify},
-    BoundedVec,
+    BoundedVec, WeakBoundedVec,
 };
 use sp_std::prelude::*;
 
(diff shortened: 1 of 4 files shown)


ivan-cholakov (approved):

nahuseyoum (approved):

pallets/nft-manager/src/lib.rs
```diff
@@ -1284,6 +1291,68 @@ impl<T: Config> InnerCallValidator for Pallet<T> {
 
 const STORAGE_VERSION: StorageVersion = StorageVersion::new(4);
 
+pub mod migrations {
```
nahuseyoum, line 1294: Is the plan to run this migration and then revert the weakbounded change?

thadouk: yes, exactly that.
That is why I have linked this PR with #194 which does exactly that.

PR #196, opened 2023-07-31
removed check for avt_lowered event

ivan-cholakov:
fixed token-manager benchmarks

### DIFF
--- a/pallets/token-manager/src/benchmarking.rs
+++ b/pallets/token-manager/src/benchmarking.rs
@@ -240,13 +240,7 @@ benchmarks! {
     )
     verify {
         assert_eq!(<T as pallet::Config>::Currency::free_balance(&lower.from_account_id), 0u32.into());
-        assert_last_event::<T>(Event::<T>::TokenLowered {
-            token_id: AVT_TOKEN_CONTRACT.into(),
-            sender: lower.from_account_id.clone(),
-            recipient: lower.lower_account_id,
-            amount: lower.amount.into(),
-            t1_recipient: lower.t1_recipient
-        }.into());
+
     }
 
     lower_non_avt_token {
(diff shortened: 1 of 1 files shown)


nahuseyoum (approved):

thadouk (approved):

pallets/token-manager/src/benchmarking.rs
```diff
@@ -240,13 +240,7 @@ benchmarks! {
...
     verify {
         assert_eq!(<T as pallet::Config>::Currency::free_balance(&lower.from_account_id), 0u32.into());
-        assert_last_event::<T>(Event::<T>::TokenLowered {
-            token_id: AVT_TOKEN_CONTRACT.into(),
-            sender: lower.from_account_id.clone(),
-            recipient: lower.lower_account_id,
-            amount: lower.amount.into(),
-            t1_recipient: lower.t1_recipient
-        }.into());
+
```
alexpArtos, line 243: Ok, so we removed the AVT `TokenLowered` event. But should we not replace it by a corresponding Lower event, from some other pallet?

thadouk: I think we should not check for output from other pallets during benchmark tests, as it goes beyond the scope of the pallet's responsibility.
Benchmarks are intended to measure the pallet's individual contribution to the overall weights. When combined with a full runtime, a runtime benchmark run will determine the full weight of the transaction for that runtime.

nahuseyoum: I don't think the verification is not linked to the benchmark code or the total weight. I think its ok to verify the effect of calling the extrinsic by checking state in another pallet.

alexpArtos: Bit of a moot point since it has been merged, but we do use the `verify` block in these benchmarks and do some basic tests in there, don't we? That is about checking the logic was correctly performed, not about checking the weight, I believe. Having more or less code in this block should not affect the benchmark measurement, but ensure that the calculations used for that measurement have been meaningful.

PR #198, opened 2023-08-02
Sys 3338 convert parachain staking to bounded vec storage items

ivan-cholakov:
Converted `pallet-parachain-staking` to use BoundedVec and implement MaxEncodedLen

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -134,7 +134,7 @@ pub mod pallet {
     /// Pallet for parachain staking
     #[pallet::pallet]
     #[pallet::storage_version(crate::migration::STORAGE_VERSION)]
-    #[pallet::without_storage_info]
+    // #[pallet::without_storage_info]
     pub struct Pallet<T>(PhantomData<T>);
 
     pub type EraIndex = u32;
@@ -181,7 +181,7 @@ pub mod pallet {
         type MaxBottomNominationsPerCandidate: Get<u32>;
         /// Maximum nominations per nominator
         #[pallet::constant]
-        type MaxNominationsPerNominator: Get<u32>;
+        type MaxNominationsPerNominator: Get<u32> + Clone + TypeInfo;
         /// Minimum stake, per collator, that must be maintained by an account that is nominating
         #[pallet::constant]
         type MinNominationPerCollator: Get<BalanceOf<Self>>;
@@ -464,7 +464,7 @@ pub mod pallet {
         _,
         Twox64Concat,
         T::AccountId,
-        Nominator<T::AccountId, BalanceOf<T>>,
+        Nominator<T::AccountId, BalanceOf<T>, T::MaxNominationsPerNominator>,
         OptionQuery,
     >;
 
@@ -1682,10 +1682,9 @@ pub mod pallet {
         pub(crate) fn update_active(candidate: T::AccountId, total: BalanceOf<T>) {
             let mut candidates = <CandidatePool<T>>::get();
             candidates.remove(&Bond::from_owner(candidate.clone()));
-            let maybe_inserted_candidate = candidates
+            candidates
                 .try_insert(Bond { owner: candidate, amount: total })
                 .map_err(|_| Error::<T>::CandidateLimitReached);
-            // candidates.insert(Bond { owner: candidate, amount: total });
             <CandidatePool<T>>::put(candidates);
         }
 
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -966,10 +966,7 @@ pub mod pallet {
                 candidates.0.len() as u32 <= candidate_count,
                 Error::<T>::TooLowCandidateCountWeightHintCancelLeaveCandidates
             );
-            // let maybe_inserted_candidate = candidates
-            //     .try_insert(Bond { owner: collator.clone(), amount: state.total_counted })
-            //     .map_err(|_| Error::<T>::CandidateLimitReached)?;
-            // ensure!(maybe_inserted_candidate, Error::<T>::AlreadyActive);
+
             match candidates
                 .try_insert(Bond { owner: collator.clone(), amount: state.total_counted })
             {
@@ -1695,10 +1692,8 @@ pub mod pallet {
         /// Caller must ensure candidate is active before calling
         pub(crate) fn update_active(candidate: T::AccountId, total: BalanceOf<T>) {
             let mut candidates = <CandidatePool<T>>::get();
-            candidates.remove(&Bond::from_owner(candidate.clone())); // TODO
-                                                                     // candidates
-                                                                     //     .try_insert(Bond { owner: candidate, amount: total })
-                                                                     //     .map_err(|_| Error::<T>::CandidateLimitReached);
+            candidates.remove(&Bond::from_owner(candidate.clone()));
+
             match candidates.try_insert(Bond { owner: candidate, amount: total }) {
                 Ok(true) => Ok(()),
                 Ok(false) => Err(Error::<T>::CandidateLimitReached),
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -100,6 +100,7 @@ pub use types::*;
 
 #[pallet]
 pub mod pallet {
+    use crate::set::BoundedOrderedSet;
     pub use crate::{
         calls::*,
         nomination_requests::{CancelledScheduledRequest, NominationAction, ScheduledRequest},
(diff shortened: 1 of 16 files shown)


alexpArtos (approved):

MBrozhko34 (approved):

pallets/parachain-staking/src/lib.rs
```diff
@@ -797,10 +807,10 @@ pub mod pallet {
...
                 Error::<T>::TooLowCandidateCountWeightHintJoinCandidates
             );
-            ensure!(
-                candidates.insert(Bond { owner: acc.clone(), amount: bond }),
-                Error::<T>::CandidateExists
-            );
+            let maybe_inserted_candidate = candidates
+                .try_insert(Bond { owner: acc.clone(), amount: bond })
+                .map_err(|_| Error::<T>::CandidateLimitReached)?;
+            ensure!(maybe_inserted_candidate, Error::<T>::CandidateExists);
```
thadouk, line 813: Is it possible to get a CanidateExists error if we have mapped it with CandidateLimitReached?
Would a match here be better to handle different error cases and drop the `ensure!` check?

pallets/parachain-staking/src/lib.rs
```diff
@@ -950,10 +960,10 @@ pub mod pallet {
...
                 Error::<T>::TooLowCandidateCountWeightHintCancelLeaveCandidates
             );
-            ensure!(
-                candidates.insert(Bond { owner: collator.clone(), amount: state.total_counted }),
-                Error::<T>::AlreadyActive
-            );
+            let maybe_inserted_candidate = candidates
+                .try_insert(Bond { owner: collator.clone(), amount: state.total_counted })
+                .map_err(|_| Error::<T>::CandidateLimitReached)?;
+            ensure!(maybe_inserted_candidate, Error::<T>::AlreadyActive);
```
thadouk, line 966: Similar for AlreadyActive errors.

pallets/parachain-staking/src/set.rs
```diff
@@ -87,3 +88,84 @@ impl<T: Ord> From<Vec<T>> for OrderedSet<T> {
...
+    fn from(v: BoundedVec<T, S>) -> Self {
+        Self::from(v)
+    }
+}
+
+// impl<T: Clone, S: Get<u32>> Clone for BoundedOrderedSet<T, S> {
+//     fn clone(&self) -> Self {
+//         BoundedOrderedSet(self.0.clone())
+//     }
+// }
```
thadouk, line 171:
```suggestion
```

pallets/parachain-staking/src/set.rs
```diff
@@ -87,3 +88,84 @@ impl<T: Ord> From<Vec<T>> for OrderedSet<T> {
...
+        v.sort();
+        let v = v.try_mutate(|inner| inner.dedup()).expect(
+            "input is a valid BoundedVec and deduping can only reduce the number of entires; qed",
+        );
+        Self::from_sorted_set(v)
+    }
+
+    /// Create a set from a `Vec`.
+    /// Assume `v` is sorted and contain unique elements.
+    pub fn from_sorted_set(v: BoundedVec<T, S>) -> Self {
```
thadouk, line 122: I know this mimics the interface of the OrderedSet, but why can't we use from instead? Is it for speed?
This function doesn't "honor" its name, it uses a bounded vector not a set as an input.
It feels prone to error especially as it depends on the input to be unique.

pallets/parachain-staking/src/set.rs
```diff
@@ -87,3 +88,84 @@ impl<T: Ord> From<Vec<T>> for OrderedSet<T> {
         Self::from(v)
     }
 }
+
+/// An ordered set backed by `BoundedVec`
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
thadouk, line 93: nice work with this data structure. I think it would be nice to have some unit tests for its interfaces.

pallets/validators-manager/src/tests/mock.rs
```diff
@@ -405,6 +405,7 @@ parameter_types! {
     pub const ErasPerGrowthPeriod: u32 = 2;
     pub const RewardPaymentDelay: u32 = 2;
     pub const RewardPotId: PalletId = PalletId(*b"av/vamgr");
+    pub const MaxCandidates: u32 = 257;
```
thadouk, line 408: why 257?

pallets/parachain-staking/src/types.rs
```diff
@@ -34,10 +34,22 @@ use sp_std::{cmp::Ordering, prelude::*};
 
 pub struct CountedNominations<T: Config> {
     pub uncounted_stake: BalanceOf<T>,
-    pub rewardable_nominations: Vec<Bond<T::AccountId, BalanceOf<T>>>,
+    pub rewardable_nominations: BoundedVec<Bond<T::AccountId, BalanceOf<T>>, MaxNominations>,
 }
 
-#[derive(Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+#[derive(Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct MaxCloneableNominations;
```
thadouk, line 41: why do we need this? Couldn't we use a ConstU32 instead?

pallets/parachain-staking/src/set.rs
ivan-cholakov: we can use `from` just a remnant of an experiment I did , it will be removed

pallets/validators-manager/src/tests/mock.rs
ivan-cholakov: because of maximum_collators_is_reached limit, so that the tests don't clash with the MaxCandidates

pallets/parachain-staking/src/types.rs
ivan-cholakov: to pass as a bound to `BoundedOrderedSet`

pallets/validators-manager/src/tests/mock.rs
alexpArtos: I'm out of the loop here, but I'd like to hear the continuation here :grin: 
It feels like this hints at some hidden dependency between variables / constants in different places. It would be good if the code could make them visible somehow, instead of using specific magic numbers without explanation.

ivan-cholakov: this is irrelevant, after the refactor of the `maybe_inserted` variables in the `lib.rs` the number of these candidates in the `validators-manager` pallet is absolutely inconsequential

pallets/parachain-staking/src/lib.rs
```diff
@@ -960,10 +966,17 @@ pub mod pallet {
                 candidates.0.len() as u32 <= candidate_count,
                 Error::<T>::TooLowCandidateCountWeightHintCancelLeaveCandidates
             );
-            let maybe_inserted_candidate = candidates
+            // let maybe_inserted_candidate = candidates
+            //     .try_insert(Bond { owner: collator.clone(), amount: state.total_counted })
+            //     .map_err(|_| Error::<T>::CandidateLimitReached)?;
+            // ensure!(maybe_inserted_candidate, Error::<T>::AlreadyActive);
+            match candidates
```
thadouk, line 973: cleanup needed

pallets/parachain-staking/src/lib.rs
```diff
@@ -1682,10 +1695,15 @@ pub mod pallet {
...
         pub(crate) fn update_active(candidate: T::AccountId, total: BalanceOf<T>) {
             let mut candidates = <CandidatePool<T>>::get();
-            candidates.remove(&Bond::from_owner(candidate.clone()));
-            candidates
-                .try_insert(Bond { owner: candidate, amount: total })
-                .map_err(|_| Error::<T>::CandidateLimitReached);
+            candidates.remove(&Bond::from_owner(candidate.clone())); // TODO
+                                                                     // candidates
+                                                                     //     .try_insert(Bond { owner: candidate, amount: total })
+                                                                     //     .map_err(|_| Error::<T>::CandidateLimitReached);
```
thadouk, line 1701: cleanup needed

pallets/validators-manager/src/tests/mock.rs
```diff
@@ -405,7 +405,7 @@ parameter_types! {
     pub const ErasPerGrowthPeriod: u32 = 2;
     pub const RewardPaymentDelay: u32 = 2;
     pub const RewardPotId: PalletId = PalletId(*b"av/vamgr");
-    pub const MaxCandidates: u32 = 257;
+    pub const MaxCandidates: u32 = 253;
```
thadouk, line 408: this magic number is changed. Now it is less than the limit on Validators Manager

runtime/test/src/lib.rs
```diff
@@ -483,6 +483,7 @@ impl pallet_parachain_staking::Config for Runtime {
     type CollatorSessionRegistration = Session;
     type CollatorPayoutDustHandler = TokenManager;
     type WeightInfo = pallet_parachain_staking::weights::SubstrateWeight<Runtime>;
+    type MaxCandidates = ConstU32<100>;
```
thadouk, line 486: I don't understand the value of this limit. How many candidates can we have? Why is 100 safe?

thadouk (approved):

PR #215, opened 2023-08-14
SYS-3556 Convert ethereum-transactions pallet to use BoundedVec

fluorostani:
## Proposed changes

converted ethereum-transactions-pallet to use BoundedVec for the number of its validators

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/pallets/ethereum-transactions/src/ethereum_transaction.rs
+++ b/pallets/ethereum-transactions/src/ethereum_transaction.rs
@@ -1,6 +1,6 @@
 use codec::{Decode, Encode, MaxEncodedLen};
 use sp_avn_common::EthTransaction;
-use sp_core::{ecdsa, H160, H256, H512, U256, ConstU32};
+use sp_core::{ecdsa, ConstU32, H160, H256, H512, U256};
 
 #[cfg(not(feature = "std"))]
 extern crate alloc;
--- a/pallets/ethereum-transactions/src/ethereum_transaction.rs
+++ b/pallets/ethereum-transactions/src/ethereum_transaction.rs
@@ -1,6 +1,6 @@
 use codec::{Decode, Encode, MaxEncodedLen};
 use sp_avn_common::EthTransaction;
-use sp_core::{ecdsa, H160, H256, H512, U256};
+use sp_core::{ecdsa, H160, H256, H512, U256, ConstU32};
 
 #[cfg(not(feature = "std"))]
 extern crate alloc;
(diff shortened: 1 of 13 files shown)


ivan-cholakov (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -424,6 +426,10 @@ pub struct EthSignatures {
 }
 
 impl EthSignatures {
+    fn max_encoded_len() -> usize {
+        u32::MAX as usize
+    }
```
thadouk, line 431: I don't think this is correct. This is the definition of the struct
```
pub struct EthSignatures {
    pub signatures_list: Vec<ecdsa::Signature>,
}
```
and `u32::MAX as usize` doesn't seem plausible for a bound of the encoded data.

In general we shouldn't implement `max_encoded_len` unless we have to, usually if you use bounded data structures all you need is to derive `MaxEncodedLen` as it was done above.
I think we should make the `signatures_list` a bounded data structure itself instead.
The good news is that the list of signatures can use the same bound as the bound we have for validators.

thadouk (changes requested):

pallets/ethereum-transactions/src/ethereum_transaction.rs
alexpArtos: So, I was doing some checks on this, just if it helps. If we _absolutely want to compute_ this maxLen, we still can.
The BoundedVec is still SCALE-encoded as a dynamic vector, which means it will only list the actual elements of the vector, and will not create empty elements for those remaining slots the vector allows for.

But it also means that the first few bytes of the vector encoding specify how many elements actually are in the vector. And this length encoding has some specific size. [Details here](https://docs.substrate.io/reference/scale-codec/#fn-1).

If the vector's length is up to 63, the prefix will be 1-byte long. If  it is up to 16,383 it will be 2-bytes long. If it is up to `2^30 - 1`, it will be 4 bytes long. And beyond that, I don't think we need to worry.

So the actual MAX_LEN would be `2 + element_size * number_of_elements` if the bound is below 16,384, or `4 + element_size * number_of_elements` if it is above.

alexpArtos (approved):

pallets/ethereum-transactions/src/ethereum_transaction.rs
alexpArtos: Just noticed that we'll probably be below 64, in which case the prefix is just `1 byte`.

alexpArtos (approved):
Generally, I want to approve this PR. Just want to be sure if the necessary migration fits in one block or we need to use the technique to restart it periodically

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -429,7 +434,7 @@ impl EthSignatures {
     }
 
     pub fn append(&mut self, mut other: Vec<ecdsa::Signature>) {
-        self.signatures_list.append(&mut other);
+        if let Err(_) = self.signatures_list.try_append(&mut other) {}
     }
```
thadouk, line 438: I think we shouldn't silently consume the error.
Normally I would suggest that we bubble up the error, but since this function is not called in the runtime, I would suggest we just delete it.

thadouk (approved):

PR #217, opened 2023-08-16
converted avn-pallet to use boundedvec

ivan-cholakov:
## Proposed changes

converted avn-pallet to use BoundedVec for the number of its validators

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ lto = "fat"
 codegen-units = 1
 
 [workspace.package]
-version = "3.2.2"
+version = "3.2.3"
 authors = ["Aventus systems team"]
 homepage = "https://www.aventus.io/"
 repository = "https://github.com/Aventus-Network-Services/avn-node-parachain/"
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -412,9 +412,7 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
             if T::DisabledValidatorChecker::is_disabled(x.0) {
                 disabled_avn_validators.push(x.0.clone());
             } else {
-                active_avn_validators
-                    .try_push(Validator::new(x.0.clone(), x.1))
-                    .map_err(|_| DispatchError::Other("Maximum Validators reached"));
+                if let Err(_) = active_avn_validators.try_push(Validator::new(x.0.clone(), x.1)) {}
             }
         });
 
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -417,9 +417,11 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
             }
         });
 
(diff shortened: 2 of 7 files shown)


pallets/avn/src/lib.rs
```diff
@@ -398,13 +406,15 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
...
+        let mut active_avn_validators = BoundedVec::default();
 
         validators.for_each(|x| {
             if T::DisabledValidatorChecker::is_disabled(x.0) {
                 disabled_avn_validators.push(x.0.clone());
             } else {
-                active_avn_validators.push(Validator::new(x.0.clone(), x.1));
+                active_avn_validators
+                    .try_push(Validator::new(x.0.clone(), x.1))
+                    .map_err(|_| DispatchError::Other("Maximum Validators reached"));
```
thadouk, line 417: we ignore the error here

Cargo.toml
```diff
@@ -12,7 +12,7 @@ lto = "fat"
 codegen-units = 1
 
 [workspace.package]
-version = "3.1.2"
+version = "3.1.3"
```
alexpArtos, line 15: This likely needs to be updated to a different number, since we're going with the `3.2.x` scheme I believe.

alexpArtos (approved):
Approving just not to put another blocker, but I'd follow Thanos's suggestions.

pallets/avn/src/lib.rs
```diff
@@ -398,13 +407,15 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
...
+        let mut active_avn_validators = WeakBoundedVec::default();
 
         validators.for_each(|x| {
             if T::DisabledValidatorChecker::is_disabled(x.0) {
                 disabled_avn_validators.push(x.0.clone());
             } else {
-                active_avn_validators.push(Validator::new(x.0.clone(), x.1));
+                if let Err(_) = active_avn_validators.try_push(Validator::new(x.0.clone(), x.1)) {
+                    return // breaking out of the for_each
+                }
```
thadouk, line 418: I think we should make this an infallible path. `active_avn_validators` can remain a normal vector and then on lone 423, we create a WeakBoundedVec with force_from and put it into storage.

pallets/avn/src/lib.rs
```diff
@@ -408,8 +417,10 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
             }
         });
 
+        let bounded_active_avn_validators = WeakBoundedVec::force_from(active_avn_validators);
```
thadouk, line 420: This could go inside the if in line 422, to optimise a little in case the list of collator hasn't changed (vast majority of the cases)

thadouk (approved):

PR #220, opened 2023-08-17
SYS-3338 Finalises NFT manager data migration to bounded data structures

thadouk:
## Proposed changes
[Diff with 3.1.1 release](https://github.com/Aventus-Network-Services/avn-parachain/compare/v3.1.1...656507f1e686fd27d736d7aa20b08dcb441756f0)
Reintroduces commit b4dfc4249bee3f50f5d73c9b5b7ea1910224e684 which finalises the NFT manager data migration.
Revert the stopgap solution for the migration and the migration code itself.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


alexpArtos:
I don't see any changes in `cargo.toml` but the diff with 3.1.1 throws this up: shouldn't we be upgrading the version here to `3.2.1` instead of `3.2.0`?

alexpArtos (approved):
Approved, but with comments on the main thread (not the code)

ivan-cholakov (approved):

thadouk:
> I don't see any changes in `cargo.toml` but the diff with 3.1.1 throws this up: shouldn't we be upgrading the version here to `3.2.1` instead of `3.2.0`?
> 
> And does the spec_version need to stay the same as well?

You are right. It was updated when the release was enabled.

pallets/nft-manager/src/lib.rs
```diff
@@ -774,37 +771,6 @@ pub mod pallet {
 
             Ok(())
         }
-
-        #[pallet::weight(T::DbWeight::get().reads_writes(1 as u64, 2 as u64))]
```
thadouk, line 778: TODO:
Ensure this is completely safe and that the call_index can be reused. If not, find out how to "burn it"

MBrozhko34 (approved):

alexpArtos (approved):

pallets/nft-manager/src/lib.rs
thadouk: Based on [this discussion on stackexchange](https://substrate.stackexchange.com/questions/8923/safe-way-to-remove-an-extrinsic):
We can remove the extrinsic and reuse the index provided we give enough time to clients to upgrade. It is safe though in the blockchain context.

PR #224, opened 2023-08-22
SYS-3338 Uses WeakBoundedVector for external reference

thadouk:
## Proposed changes

Removes the migration of data during the upgrade since for large numbers of NFTs the migration could hault the upgrade.
Unique external reference field will use WeakBoundedVector to ensure that storage can be decoded for operations.

## Type of change/Merge

- [x] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/pallets/nft-manager/src/lib.rs
+++ b/pallets/nft-manager/src/lib.rs
@@ -86,6 +86,8 @@ pub type NftUniqueId = U256;
 /// Suggested bound to use in runtime for number of NFTs that can exist in a single Batch
 pub type BatchNftBound = ConstU32<16384>;
 
+const STORAGE_VERSION: StorageVersion = StorageVersion::new(4);
+
 #[frame_support::pallet]
 pub mod pallet {
     use super::*;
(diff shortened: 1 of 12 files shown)


ivan-cholakov (approved):

alexpArtos (approved):
I wonder if it wouldn't make sense to use a limit on the actual removal of storage. That would protect us against big changes, apparently, but would require us to probably redo the migration several times.

fluorostani (approved):

ivan-cholakov (approved):

pallets/nft-manager/src/lib.rs
```diff
@@ -789,65 +791,34 @@ pub mod pallet {
...
                     Pallet::<T>::on_chain_storage_version()
                 );
 
-                return migrations::migrate_to_bounded_nft::<T>()
+                return migration_weight
             }
             return Weight::from_ref_time(0)
         }
+
+        fn on_initialize(_n: T::BlockNumber) -> Weight {
```
ivan-cholakov, line 809: what is this hook? how is this different from `on_runtime_upgrade`

thadouk: This is a hook that is called on the initialisation of each block, allowing you to perform actions on system-reserved space for each block

alexpArtos: I suspect that `on_runtime_upgrade` will run once, when an upgrade is made, but `on_initialize` will run on any new block?

PR #225, opened 2023-08-24
SYS-3630 Update Quorum from 2/3 to 1/3

fluorostani:


pallets/ethereum-transactions/src/tests/tests.rs
```diff
@@ -172,13 +172,13 @@ impl DefaultTransactionBuilder {
         return [3u8; 32]
     }
 
-    // The quorum is 2/3 of the validators plus 1
-    // Our validator list has 3 members, so the quorum must be 3
+    // The quorum is 1/3 of the validators
+    // Our validator list has 3 members, so the quorum must be 1
     // I chose not to implement the same logic of the quorum function
```
alexpArtos, line 177: :100:

pallets/ethereum-transactions/src/tests/tests.rs
```diff
@@ -172,13 +172,13 @@ impl DefaultTransactionBuilder {
...
     // because that would defeat the point of the test being black box
     pub fn default_quorum() -> u32 {
```
alexpArtos, line 179: I think we should have a test for more values of validators. We should test at least for n=4, n=10, and maybe the transitions between sets of 3, eg, 6 and 7.

ivan-cholakov (approved):

pallets/summary/src/tests/mock.rs
```diff
@@ -567,7 +570,10 @@ pub const FIRST_VALIDATOR_INDEX: u64 = 1;
 pub const SECOND_VALIDATOR_INDEX: u64 = 2;
 pub const THIRD_VALIDATOR_INDEX: u64 = 3;
 pub const FOURTH_VALIDATOR_INDEX: u64 = 4;
-pub const VALIDATORS_COUNT: u64 = 4;
+pub const FIFTH_VALIDATOR_INDEX: u64 = 5;
```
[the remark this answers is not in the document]
fluorostani: We do, it depends upon the test but most get added as ranges

alexpArtos (approved):

alexpArtos (approved):

pallets/avn/src/lib.rs
```diff
@@ -296,13 +361,9 @@ impl<T: Config> Pallet<T> {
...
     }
 
-    pub fn calculate_two_third_quorum() -> u32 {
-        let len = Self::active_validators().len() as u32;
-        if len < 3 {
-            return len
-        } else {
-            return (2 * len / 3) + 1
-        }
+    pub fn calculate_one_third_quorum() -> u32 {
```
nahuseyoum, line 364: we seem to have this function defined twice. Once in Primitives and another here.

nahuseyoum (commented):
changes LGTM but I think we should change the base on this PR to be a feature branch.

pallets/avn/src/lib.rs
fluorostani: Removed from the avn pallet, everything references the primitives implementation now

alexpArtos (approved):

alexpArtos (approved):

pallets/avn-finality-tracker/src/mock.rs
```diff
@@ -73,12 +73,21 @@ impl system::Config for TestRuntime {
     type MaxConsumers = frame_support::traits::ConstU32<16>;
 }
 
+impl pallet_timestamp::Config for TestRuntime {
```
nahuseyoum, line 76: To avoid adding `pallet_timestamp` as a dependency everywhere and implement it in mock.rs, you can implement `UnixTime` in avn and use that in your mock.rs files. So something like this should work:

- In avn, lib.rs
```
pub struct TestUnixTime {
}
impl UnixTime for TestUnixTime  {
    fn now() -> core::time::Duration {
      core::time::Duration::from_millis(0u64)
    }
}
```

- In your mock.rs files:
```
type TimeProvider = pallet_avn::TestUnixTime;
```

PR #227, opened 2023-08-29
Sys 3638 remove sender from signature

ivan-cholakov:
This is a feature branch pr for removing sender id from signature


pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -306,16 +306,10 @@ impl EthAbiHelper {
     }
 
     pub fn generate_ethereum_description_for_signature_request(
```
alexpArtos, line 308: Probably rename this function. I think what it is actually doing is `encode_arguments`.

pallets/ethereum-transactions/src/tests/ethereum_transaction_tests.rs
```diff
@@ -203,7 +201,7 @@ fn short_ethereum_description_appends_to_original_description() {
     assert_eq!(&inputs[0..number_of_original_terms], &original_inputs[..]);
     assert_eq!(&values[0..number_of_original_terms], &original_values[..]);
 
-    let expected_new_arguments = ["_t2TransactionId", "_t2PublicKey"];
+    let expected_new_arguments = ["_t2TransactionId"];
```
alexpArtos, line 204: :+1:

pallets/summary/src/lib.rs
```diff
@@ -880,29 +879,16 @@ pub mod pallet {
...
-                .map_err(|_| Error::<T>::InvalidRoot)?;
+            let maybe_root_data_tx_id = match root_data.tx_id {
+                None => EMPTY_ROOT_TRANSACTION_ID,
+                _ => *root_data
+                    .tx_id
+                    .as_ref()
+                    .expect("Non-Empty roots have a reserved TransactionId"),
+            };
 
             Ok(hex::encode(EthAbiHelper::generate_eth_abi_encoding_for_params_only(
```
alexpArtos, line 890: I think this name is not accurate anymore. All we want is to create an encoded payload, an encoded series of bytes that represent some data relative to the transaction. It may be some of the arguments, it may be all, they may even be in a different order, so this is not really an encoding of the params of the Tx. This name does not help in understanding what the code does.

alexpArtos: And on that note, if all we are doing here is encoding directly the data, and we have removed the description (because the rest of the information in there, namely the `inputs`, is not necessary) then we have a simpler task. We can create the bytes array here, push the necessary data into it and encode it. It is all at the same level of abstraction and because of that does not necessarily need to call into another function.

Currently, all we are returning here is the TxId and the Root Hash. We will want to add the Expiry when that work is done, but if these three things are passed from the outside, there is no need to call another function to push them into the array before encoding.

pallets/ethereum-transactions/src/ethereum_transaction.rs
```diff
@@ -335,8 +329,10 @@ impl EthAbiHelper {
         Ok(mut_call)
     }
 
-    pub fn generate_eth_abi_encoding_for_params_only(call: &EthTransactionDescription) -> Vec<u8> {
-        ethabi::encode(&call.call_values)
+    pub fn generate_eth_abi_encoding_for_params_only(transaction_id: TransactionId) -> Vec<u8> {
```
alexpArtos, line 332: we probably don't need this function.

pallets/ethereum-transactions/src/ethereum_transaction.rs
ivan-cholakov: This function is no longer used anywhere, so it is to be deleted

alexpArtos (approved):

pallets/avn/src/lib.rs
```diff
@@ -146,6 +200,18 @@ impl<T: Config> Pallet<T> {
         Ok(maybe_validator.expect("Already checked"))
     }
 
+    pub fn calculate_eth_tx_expiry() -> U256 {
+        let now = T::TimeProvider::now();
+        let now_as_secs = now.as_secs();
+
+        let timeout_duration_hours = Self::get_eth_tx_timeout_duration();
+        let timeout_duration_seconds = timeout_duration_hours * 3600;
```
fluorostani, line 208: Could we make the storage value seconds instead? This would afford more fine-grained control

pallets/avn/src/lib.rs
```diff
@@ -110,6 +122,48 @@ pub mod pallet {
     /// offchain worker.
     pub type Validators<T: Config> =
         StorageValue<_, Vec<Validator<T::AuthorityId, T::AccountId>>, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn get_eth_tx_timeout_duration)]
+    pub type TimoutDuration<T: Config> = StorageValue<_, u64, ValueQuery>;
```
fluorostani, line 128:
```suggestion
    pub type TimeoutDuration<T: Config> = StorageValue<_, u64, ValueQuery>;
```

pallets/avn/src/lib.rs
```diff
@@ -110,6 +122,48 @@ pub mod pallet {
...
+    impl<T: Config> Default for GenesisConfig<T> {
+        fn default() -> Self {
+            Self { _phantom: Default::default(), timeout_duration: 0 }
+        }
+    }
+
+    #[pallet::genesis_build]
+    impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+        fn build(&self) {
+            TimoutDuration::<T>::put(self.timeout_duration);
```
fluorostani, line 146:
```suggestion
            TimeoutDuration::<T>::put(self.timeout_duration);
```

pallets/avn/src/lib.rs
```diff
@@ -110,6 +122,48 @@ pub mod pallet {
...
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::set_eth_tx_timeout_duration())]
+        pub fn set_eth_tx_timeout_duration(
+            origin: OriginFor<T>,
+            timeout_duration: u64,
+        ) -> DispatchResultWithPostInfo {
+            // Ensure that the origin is the sudo key
+            ensure_root(origin)?;
+
+            // Update the timeout_duration storage item
+            TimoutDuration::<T>::put(timeout_duration);
```
fluorostani, line 162:
```suggestion
            TimeoutDuration::<T>::put(timeout_duration);
```

pallets/avn/src/lib.rs
ivan-cholakov: sure will remove the calculation and will pass the value of the storage item directly

fluorostani (approved):

alexpArtos (approved):

PR #229, opened 2023-08-30
added a sudo action to set a timeout duration

ivan-cholakov:


pallets/avn/src/lib.rs
```diff
@@ -201,9 +201,9 @@ impl<T: Config> Pallet<T> {
     }
 
     pub fn calculate_eth_tx_expiry() -> U256 {
-        // let now = T::TimeProvider::now();
-        // let now_as_secs = now.as_secs();
-        let now_as_secs = 123;
+        let now = T::TimeProvider::now();
+        let now_as_secs = now.as_secs();
+        // let now_as_secs = 123;
```
alexpArtos, line 206: can we remove this line please?

alexpArtos (approved):

alexpArtos:
What is the purpose of this PR? Why was it merged without any approval? That defeats the point of having PRs, even if it is for a feature-branch.

The point I need to understand is why we are adding an extrinsic to set timeout. Was this discussed anywhere?

PR #230, opened 2023-09-05
SYS-3611: updated benchmarks using avn pallet

ivan-cholakov:
## Proposed changes

Updated benchmarking tests, related to avn-pallet

## Type of change/Merge

- [x] Release 
  - [ ] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


thadouk (approved):
I would remove the runtime version bump and minor release and let the next release consume this, as it is only Benchmarks.

PR #231, opened 2023-09-05
DEVOPS-457 fix PR workflows on anv-parachain repo

vukomir:
## Proposed changes

fixed Runtime Benchmarks on PR workflow 
## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -101,6 +101,10 @@ jobs:
     timeout-minutes: 90
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: ${{ inputs.fetch-depth }}
+          ref: ${{ inputs.ref }}
+          repository: ${{ inputs.repository }}
 
       - name: Install Rust
         uses: actions-rs/toolchain@v1


ivan-cholakov (approved):

RuiSMagalhaes (approved):

thadouk (approved):

.github/workflows/cargo-test.yml
```diff
@@ -101,6 +101,10 @@ jobs:
     timeout-minutes: 90
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: ${{ inputs.fetch-depth }}
+          ref: ${{ inputs.ref }}
+          repository: ${{ inputs.repository }}
```
thadouk, line 107: Why do we need to do a special checkout for benchmarks?
Shouldn't all the tests check out the latest code? For example the test workflow on the same file doesn't do the same checkout as the benchmarks although they both run on the same PR

ivan-cholakov: look at line 63

PR #238, opened 2023-09-19
SYS-3663 Add voting infrastructure

nahuseyoum:


pallets/parachain-staking/src/vote.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl<T: Config> VotingSessionManager<T::AccountId, T::BlockNumber> for GrowthVotingSession<T> {
+    fn cast_vote_context(&self) -> &'static [u8] {
+        return CAST_VOTE_CONTEXT
+    }
+
+    fn end_voting_period_context(&self) -> &'static [u8] {
+        return END_VOTING_PERIOD_CONTEXT
+    }
```
alexpArtos, line 53: do we need to have functions merely to access a constant? Can we not  use the constant directly?

pallets/parachain-staking/src/vote.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    fn state(&self) -> Result<VotingSessionData<T::AccountId, T::BlockNumber>, DispatchError> {
+        if <ParachainStaking<T> as Store>::VotesRepository::contains_key(self.growth_id) {
+            return Ok(ParachainStaking::<T>::get_vote(self.growth_id))
+        }
+        return Err(DispatchError::Other("Growth data is not found in votes repository"))
+    }
+
+    fn is_valid(&self) -> bool {
+        return false;
```
alexpArtos, line 63: Is this intentional?

pallets/parachain-staking/src/vote.rs
```diff
@@ -0,0 +1,223 @@
...
+                vote.confirmations
+                    .try_push(approval_signature)
+                    .map_err(|_| avn_error::<T>::VectorBoundsExceeded)?;
+                Ok(())
+            },
+        )?;
+        Ok(())
+    }
+
+    fn record_reject_vote(&self, voter: T::AccountId) -> DispatchResult {
```
alexpArtos, line 124: Why don't we use signatures with a reject vote, as we do with approvals?

alexpArtos (approved):

ivan-cholakov (approved):

pallets/parachain-staking/src/vote.rs
```diff
@@ -0,0 +1,223 @@
...
+        return ParachainStaking::<T>::end_voting(sender, &self.growth_id)
+    }
+}
+
+/***************** Functions that run in an offchain worker context  **************** */
+
+pub fn create_vote_lock_name<T: Config>(growth_id: &GrowthId) -> OcwLock::PersistentId {
+    let mut name = b"vote_growth::hash::".to_vec();
+    name.extend_from_slice(&mut growth_id.period.encode());
+    name.extend_from_slice(&mut growth_id.ingress_counter.encode());
```
fluorostani, line 145:
```suggestion
    name.extend_from_slice(&growth_id.period.encode());
    name.extend_from_slice(&growth_id.ingress_counter.encode());
```

pallets/parachain-staking/src/vote.rs
```diff
@@ -0,0 +1,223 @@
...
+use sp_core::ecdsa;
+use sp_runtime::scale_info::TypeInfo;
+use sp_std::fmt::Debug;
+
+use super::{Call, Config};
+use crate::{Pallet as ParachainStaking, GrowthId, Store, AVN, GrowthPeriodIndex};
+
+pub const CAST_VOTE_CONTEXT: &'static [u8] = b"growth_casting_vote";
+pub const END_VOTING_PERIOD_CONTEXT: &'static [u8] = b"growth_end_voting_period";
+const MAX_VOTING_SESSIONS_RETURNED: usize = 5;
```
fluorostani, line 32: Does this get used?

pallets/parachain-staking/src/vote.rs
fluorostani, line 1: General comment, do we need all the returns at the ends of functions? I thought it was implicit in Rust?

nahuseyoum: There are 2 ways to return:
 - with an explicit return and a semicolon at the end
 - without an explicit return and no semicolon - if its the last line of a function
I remember deciding to choose the more explicit version to avoid accidentally returning by simply forgetting to add a semicolon.

pallets/parachain-staking/src/vote.rs
nahuseyoum: Yes, in the next PRs

pallets/parachain-staking/src/vote.rs
nahuseyoum: Yes, this was a palceholder for now. It is implemented in the next few PRs.

pallets/parachain-staking/src/vote.rs
nahuseyoum: These functions are interface functions that need to be implemented for each pallet.

pallets/parachain-staking/src/vote.rs
nahuseyoum: Because we don't need to send anything to Ethereum if we reject a proposal. The AvN signature is required because we need to make sure a genuine collator is voting but the Eth signature is not required.

pallets/parachain-staking/src/vote.rs
alexpArtos: Safer code usually gets my vote, so I side with Nahu on this one.

PR #243, opened 2023-09-22
SYS-3640 update LogLifted event signature

MBrozhko34:
## Proposed changes

Updated LiftedData parameters and LogLifted event signature removing tier 1 address.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/primitives/avn-common/src/event_types.rs
+++ b/primitives/avn-common/src/event_types.rs
@@ -82,9 +82,9 @@ impl ValidEvents {
             ValidEvents::AddedValidator =>
                 H256(hex!("ff083a6e395a67771f3c9108922bc274c27b38b48c210b0f6a8c5f4710c0494b")),
 
-            // hex string of Keccak-256 for LogLifted(address,address,bytes32,uint256)
+            // hex string of Keccak-256 for LogLifted(address,bytes32,uint256)
             ValidEvents::Lifted =>
-                H256(hex!("8964776336bc2fa8ecaaf70b6f8e8450807efb1ff78f8b87980707aa821f0ec0")),
+                H256(hex!("418da8f85cfa851601f87634c6950491b6b8785a6445c8584f5658048d512cae")),
 
             // hex string of Keccak-256 for AvnMintTo(uint256,uint64,bytes32,string)
             ValidEvents::NftMint =>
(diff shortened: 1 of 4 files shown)


nahuseyoum (approved):

alexpArtos (commented):
Without having run the tests, I think there is at least one thing missing in this PR, and that is to update the `signature` for LogLifted around lines 85-87. This code here: 

```
            // hex string of Keccak-256 for LogLifted(address,address,bytes32,uint256)
            ValidEvents::Lifted =>
                H256(hex!("8964776336bc2fa8ecaaf70b6f8e8450807efb1ff78f8b87980707aa821f0ec0")),
```

But this triggered other thoughts:
- the current changes make any previous LogLifted event invalid, even though they still exist on the chain and were valid at that time. Wouldn't it be better to keep both versions and be able to validate the current events and the late events?
- also, shouldn't we run tests with an actual even following the new format, before submitting the PR? Without the signature change, I'm not sure we will be able to validate a new event and would like to have some confirmation of that before accepting the PR

primitives/avn-common/src/event_types.rs
```diff
@@ -208,21 +208,17 @@ impl AddedValidatorData {
...
-    pub sender_address: H160,
     pub receiver_address: H256,
     pub amount: u128,
     pub nonce: U256,
 }
 
 impl LiftedData {
     const TOPIC_CURRENCY_CONTRACT: usize = 1;
-    const TOPIC_INDEX_T1_ADDRESS: usize = 2;
     const TOPIC_INDEX_T2_ADDRESS: usize = 3;
```
alexpArtos, line 218: I think this needs to change as well, now that I go into the details. Now that we don't have the sender data in the event, we lose one of the topics, so what comes after it (the T2 address) moves up one position to be number 2.

alexpArtos, line 218:
```suggestion
    const TOPIC_INDEX_T2_ADDRESS: usize = 2;
```

alexpArtos (approved):
Conditional approval. See comments

MBrozhko34:
Working on resolving tests

alexpArtos (approved):
Well done, all the tests pass, I think we're good to go now!

PR #245, opened 2023-09-25
SYS-3663 Implement unsigned transactions support

nahuseyoum:


pallets/parachain-staking/src/lib.rs
```diff
@@ -1715,6 +1743,87 @@ pub mod pallet {
...
+
+                return approve_vote_validate_unsigned::<T>(
+                    &growth_voting_session,
+                    validator,
+                    eth_encoded_data.encode(),
+                    approval_signature,
+                    signature,
+                )
+            } else if let Call::reject_growth { growth_id, validator, signature } = call {
+                let growth_voting_session = Self::get_growth_voting_session(growth_id);
```
alexpArtos, line 1817: Why don't we need to check if the growth_id.period exists, as we did in the approve branch?

pallets/parachain-staking/src/lib.rs
```diff
@@ -2508,11 +2626,99 @@ pub mod pallet {
...
+                    growth.triggered = Some(true)
+                });
+                
+                Self::deposit_event(Event::<T>::TriggeringGrowth {
+                    growth_period: growth_id.period,                    
+                });
+            } else {
+                // We didn't get enough votes to approve this growth
+
+                let growth_creator =
```
alexpArtos, line 2686: Is this needed at the moment? It does not seem used anywhere.

alexpArtos (approved):

pallets/parachain-staking/src/lib.rs
nahuseyoum: No its not, I have another PR that cleans up unused references but will also remove it here.

pallets/parachain-staking/src/lib.rs
nahuseyoum: Good point, added extra check.

PR #247, opened 2023-09-25
SYS-3652 Trigger growth on T1

nahuseyoum:


pallets/parachain-staking/src/lib.rs
```diff
@@ -2532,27 +2539,53 @@ pub mod pallet {
...
+                    average_staked_in_period_128,
+                    *growth_period
+                ));
+
+                let tx_id =  Some(T::CandidateTransactionSubmitter::reserve_transaction_id(&trigger_growth).map_err(|_| Error::<T>::ErrorReservingId)?);                            
                 let added_by = 
                     AVN::<T>::calculate_primary_validator(<frame_system::Pallet<T>>::block_number())
                     .map_err(|_| Error::<T>::ErrorCalculatingPrimaryValidator)?;
+                
+                let quorum = calculate_two_third_quorum(AVN::<T>::validators().len() as u32);
```
alexpArtos, line 2563: We have introduced and merged PRs to reduce the quorum to 1/3. Should this not reflect that?

alexpArtos (approved):

pallets/parachain-staking/src/lib.rs
fluorostani: @alexpArtos - those are on [a feature branch](https://github.com/Aventus-Network-Services/avn-parachain/pull/225), so the growth work will use legacy methods for now. Then we can convert it along with the other author functions later.

fluorostani (approved):

PR #253, opened 2023-09-27
SYS-3634 Add eth-bridge pallet

fluorostani:
This pallet provides a single interface, **publish_to_avn_bridge**, which enables other pallets to execute any author-accessible function on the Ethereum-based **avn-bridge** contract.

Pallets do so by passing the desired avn-bridge function name, along with an array of parameter tuples, each comprising the data type and its corresponding value.

Upon receipt of a **publish_to_avn_bridge** request this pallet takes charge of the entire transaction process, culminating in the conclusive determination of the transaction's status on Ethereum. A callback is then made to the originating pallet which it can use to determine whether to commit or rollback state.

Specifically, the pallet manages:
- The packaging and encoding of the transaction to ensure Ethereum compatibility.
- The addition of a timestamp, delineating the deadline by which the transaction must reach the contract.
- The addition of a unique transaction ID, against which request data can be stored on the AvN and the transaction's status in the avn-bridge can be later checked.
- Collection of the necessary ECDSA signatures from authors, labelled **confirmations**, which serve to prove AvN consensus for the transaction to the avn-bridge.
- Appointing an author responsible for sending the transaction to Ethereum.
- Utilising the transaction ID and expiry to check the status of a sent transaction on Ethereum and arrive at a consensus of that status by providing **corroborations**.
- Alerting the originating pallet to the final outcome via the HandleAvnBridgeResult callback. 

The core of the pallet resides in the off-chain worker. The OCW monitors all unresolved transactions, prompting authors to resolve them by invoking one of three unsigned extrinsics:

1. Before a transaction can be dispatched, confirmations are accumulated from non-sending authors via the **add_confirmation** extrinsic until a consensus is reached. _Note: the sender's confirmation is implicit._
2. Once a transaction has received sufficient confirmations, the chosen sender is prompted to dispatch it to Ethereum and tag it as sent using the **add_receipt** extrinsic.
3. Finally, when a transaction possesses a receipt, or if its expiration time has elapsed without a definitive outcome, all authors except the sender are requested to **add_corroboration**s which will determine the final state.


alexpArtos (approved):

ivan-cholakov (approved):

pallets/eth-bridge/src/tests/tests.rs
```diff
@@ -0,0 +1,91 @@
...
+
+const ROOT_HASH: &str = "30b83f0d722d1d4308ab4660a72dbaf0a7392d5674eca3cd21d57256d42df7a0";
+const REWARDS: &[u8] = b"15043665996000000000";
+const AVG_STAKED: &[u8] = b"9034532443555111110000";
+const PERIOD: &[u8] = b"3";
+const T2_PUB_KEY: &str = "14aeac90dbd3573458f9e029eb2de122ee94f2f0bc5ee4b6c6c5839894f1a547";
+const T1_PUB_KEY: &str = "23d79f6492dddecb436333a5e7a4cfcc969f568e01283fa2964aae15327fb8a3b685a4d0f3ef9b3c2adb20f681dbc74b7f82c1cf8438d37f2c10e9c79591e9ea";
+
+#[test]
+fn check_publish_root_encoding() {
```
alexpArtos, line 14: Can we add a comment please explaining how we obtained the encoded calldata, in case we need to regenerate it again in the future when we change something.

alexpArtos (approved):

pallets/eth-bridge/src/tests/tests.rs
fluorostani: Tests will be changing anyway, these were just to keep sanity whilst developing really

PR #272, opened 2023-10-15
SYS-3493 Remove `avn-finality-tracker` pallet from runtime

nahuseyoum:
## Proposed changes

As a parachain, every block we produce is based on a finalised block of the relay chain and there is no need to implement a separate logic to check and enforce finality.

In addition, this change will significantly increase the speed of processing cross chain transactions and reduce block space occupied by unsigned (system) transactions.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


thadouk (approved):
It's better to keep the logic in code sections that need finality but make the implementation return 'true' in the parachain context. These libraries can also be used in a solochain where finality on chain data is needed and the avn-finality-tracker will be the implementation of that.

thadouk:
> > It's better to keep the logic in code sections that need finality but make the implementation return 'true' in the parachain context. These libraries can also be used in a solochain where finality on chain data is needed and the avn-finality-tracker will be the implementation of that.
> 
> > It's better to keep the logic in code sections that need finality but make the implementation return 'true' in the parachain context. These libraries can also be used in a solochain where finality on chain data is needed and the avn-finality-tracker will be the implementation of that.
> 
> The only point of doing that, as you said, is if we use this in a solo chain (or a chain where finality is not guaranteed). I suppose we could have a ticket for that in case we run a solo chain again.

The dependency on block finality should be captured regardless as part of the pallet code. The pallets don't care about the implementation and are agnostic on the runtime and the chain type that uses them.

nahuseyoum:
> > > It's better to keep the logic in code sections that need finality but make the implementation return 'true' in the parachain context. These libraries can also be used in a solochain where finality on chain data is needed and the avn-finality-tracker will be the implementation of that.
> > 
> > 
> > > It's better to keep the logic in code sections that need finality but make the implementation return 'true' in the parachain context. These libraries can also be used in a solochain where finality on chain data is needed and the avn-finality-tracker will be the implementation of that.
> > 
> > 
> > The only point of doing that, as you said, is if we use this in a solo chain (or a chain where finality is not guaranteed). I suppose we could have a ticket for that in case we run a solo chain again.
> 
> The dependency on block finality should be captured regardless as part of the pallet code. The pallets don't care about the implementation and are agnostic on the runtime and the chain type that uses them.

This is a good point. I agree we can have a configurable way to check for finalisation and leave it defaulting to `finalised` for our parachain.

PR #282, opened 2023-10-25
SYS-3634 Clarify naming, remove superfluous bytes check and double refs

fluorostani:


alexpArtos (approved):

pallets/eth-bridge/src/eth.rs
```diff
@@ -236,15 +241,11 @@ fn process_send_response<T: Config>(result: Vec<u8>) -> Result<H256, DispatchErr
...
+fn process_corroborate_result<T: Config>(result: Vec<u8>) -> Result<i8, DispatchError> {
+    let result_bytes = hex::decode(&result).map_err(|_| Error::<T>::InvalidBytes)?;
 
     if result_bytes.len() != 32 {
-        return Err(Error::<T>::InvalidViewResponseLength.into())
-    }
-
-    if !result_bytes[..31].iter().all(|&byte| byte == 0 || byte == 0xFF) {
-        return Err(Error::<T>::InvalidViewResponseValue.into())
+        return Err(Error::<T>::InvalidBytesLength.into())
```
nahuseyoum, line 248: is there a scenario where we allow the contract to send any random 32 bytes and only extract the last byte? if we always expect the contract to send 32 bytes where the first 31 bytes are all 0, then why not validate that?

nahuseyoum (approved):

pallets/eth-bridge/src/eth.rs
fluorostani: Ideally the contract would return the 1 byte we want. Alas, it pads to 32. The 31 padded bytes will always be either all 0 for positive or all FF for negative. But as long as we are getting the expected result type (32 bytes) then we can just use the single byte we care about. No need to double check expected Ethereum padding.

PR #286, opened 2023-10-26
SYS-3634 Fix benchmarks

nahuseyoum:


pallets/eth-bridge/src/benchmarking.rs
```diff
@@ -120,10 +130,12 @@ benchmarks! {
         let active_tx = ActiveTransaction::<T>::get().expect("is active");
         let msg_hash_string = active_tx.msg_hash.to_string();
         // TODO: We need a real signature as this benchmark fails at present
```
fluorostani, line 132:
```suggestion
```

fluorostani (approved):

ivan-cholakov (approved):

pallets/eth-bridge/src/benchmarking.rs
nahuseyoum: I think this comment is still relevant because the real benchmark will fail due to this signature. Will fix this in a new PR.

PR #295, opened 2023-11-08
Sys 3732 wireup offence pallet

ivan-cholakov:
Adding offences to eth-bridge, tests will be added in a follow-up pr

### DIFF
--- a/pallets/eth-bridge/src/offence.rs
+++ b/pallets/eth-bridge/src/offence.rs
@@ -1,0 +1,57 @@
+use codec::{Decode, Encode, MaxEncodedLen};
+use pallet_session::{historical::IdentificationTuple, Config as SessionConfig};
+use scale_info::TypeInfo;
+use sp_runtime::Perbill;
+use sp_staking::{
+    offence::{Kind, Offence},
+    SessionIndex,
+};
+
+#[derive(PartialEq, Eq, Clone, Debug, Encode, Decode, TypeInfo)]
+pub enum EthBridgeCorroborationOffenceType {
+    ChallengeAttemptedOnValidResult,
+}
+
+#[derive(PartialEq, Clone, Debug, Encode, Decode, TypeInfo)]
+pub struct EthBridgeCorroborationOffence<Offender> {
+    pub session_index: SessionIndex,
+    pub offenders: Vec<Offender>,
+    pub offence_type: EthBridgeCorroborationOffenceType,
+    pub validator_set_count: u32,
+}
+
+impl<Offender: Clone> Offence<Offender> for EthBridgeCorroborationOffence<Offender> {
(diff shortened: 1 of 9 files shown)


pallets/eth-bridge/src/offence.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    fn offenders(&self) -> Vec<Offender> {
+        self.offenders.clone()
+    }
+
+    fn session_index(&self) -> SessionIndex {
+        self.session_index
+    }
+
+    fn validator_set_count(&self) -> u32 {
```
alexpArtos, line 39: Why do we need this? Isn't the result of this function to enable us to call `obj.validator_set_count` from the outside? But if that is the same name  of the internal field, why is this not redundant?

pallets/eth-bridge/src/offence.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl<Offender: Clone> Offence<Offender> for CorroborationOffence<Offender> {
+    const ID: Kind = *b"bridge:::offence";
+    type TimeSlot = SessionIndex;
+
+    fn offenders(&self) -> Vec<Offender> {
+        self.offenders.clone()
+    }
+
+    fn session_index(&self) -> SessionIndex {
```
alexpArtos, line 35: see below

pallets/eth-bridge/src/tests/mock.rs
```diff
@@ -2,20 +2,40 @@ use super::*;
...
+type Offence = crate::CorroborationOffence<IdentificationTuple>;
+pub struct OffenceHandler;
+impl ReportOffence<AccountId, IdentificationTuple, Offence> for OffenceHandler {
+    fn report_offence(reporters: Vec<AccountId>, offence: Offence) -> Result<(), OffenceError> {
+        OFFENCES.with(|l| l.borrow_mut().push((reporters, offence)));
+        Ok(())
+    }
+
+    fn is_known_offence(_offenders: &[IdentificationTuple], _time_slot: &SessionIndex) -> bool {
+        false
```
alexpArtos, line 35: Will this change later?

alexpArtos (approved):

pallets/eth-bridge/src/offence.rs
ivan-cholakov: this is the implementation of the `Offence` trait, so this implementation is just a stub, and yes it is redundant

pallets/eth-bridge/src/tests/mock.rs
ivan-cholakov: this is just basic implementation, the tests are going to come with the `eth-bridge` unit tests pr

PR #309, opened 2023-11-22
added migration to drop storage items

ivan-cholakov:
created a migration to remove storage items that were cleaned up in code

### DIFF
--- a/pallets/validators-manager/Cargo.toml
+++ b/pallets/validators-manager/Cargo.toml
@@ -69,7 +69,7 @@ std = [
 	"pallet-avn-proxy/std",
 	"pallet-parachain-staking/std",
 	"libsecp256k1?/std",
-	"log/std",
+	# "log/std",
 ]
 runtime-benchmarks = [
     "frame-benchmarking/runtime-benchmarks",
--- a/pallets/validators-manager/Cargo.toml
+++ b/pallets/validators-manager/Cargo.toml
@@ -11,6 +11,7 @@ repository = { workspace = true }
 rust-version = { workspace = true }
 
 [dependencies]
+log = { version = "0.4.17", default-features = false }
 serde = { version = "1.0.137", optional = true }
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
 hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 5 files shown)


fluorostani (approved):

pallets/validators-manager/Cargo.toml
```diff
@@ -68,6 +69,7 @@ std = [
 	"pallet-avn-proxy/std",
 	"pallet-parachain-staking/std",
 	"libsecp256k1?/std",
+	"log/std",
```
nahuseyoum, line 72: Do we need this?

pallets/validators-manager/src/migration.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+pub fn remove_storage_items<T: Config>() -> Weight {
+    let mut consumed_weight: Weight = Weight::from_ref_time(0);
+    let mut add_weight = |reads, writes, weight: Weight| {
+        consumed_weight += T::DbWeight::get().reads_writes(reads, writes);
+        consumed_weight += weight;
+    };
+
+    storage_with_voting::PendingApprovals::<T>::drain();
```
nahuseyoum, line 72: IMO `drain` is intended to remove the values and return them so they can be iterated on. Perhaps something like this would be better? 
```
 let prefix = storage::storage_prefix(b"ValidatorsManager", b"PendingApprovals");
    let mut key = vec![0u8; 32];
    key[0..32].copy_from_slice(&prefix);
    let _ = unhashed::clear_prefix(&key[0..32], None, None);
```
Or maybe use `kill`?

nahuseyoum (approved):

pallets/validators-manager/src/migration.rs
```diff
@@ -0,0 +1,82 @@
...
+    pallet_prelude::*,
+    traits::{Get, OnRuntimeUpgrade},
+    weights::Weight, storage_alias,
+};
+use sp_avn_common::IngressCounter;
+use pallet_avn::vote::VotingSessionData;
+
+use crate::{Config, Pallet, ActionId};
+
+
```
alexpArtos, line 12:
```suggestion

```

pallets/validators-manager/src/migration.rs
```diff
@@ -0,0 +1,82 @@
...
+            log::info!(
+                "ℹ️  Validator manager data migration invoked with current storage version {:?} / onchain {:?}",
+                current,
+                onchain
+            );
+            return remove_storage_items::<T>()
+        }
+
+        
+
```
alexpArtos, line 50:
```suggestion

```

alexpArtos (approved):

pallets/validators-manager/src/migration.rs
ivan-cholakov: `kill` cannot be used in the context of a map.

PR #325, opened 2023-12-04
SYS-3775 Fix existing test and benchmark

nahuseyoum:


ivan-cholakov (approved):

alexpArtos (approved):

pallets/eth-bridge/src/benchmarking.rs
```diff
@@ -207,7 +206,7 @@ benchmarks! {
     }: _(RawOrigin::None, tx_id, new_confirmation.clone(), author.clone(), signature)
     verify {
         let active_tx = ActiveRequest::<T>::get().expect("is active");
-        ensure!(active_tx.confirmations.contains(&new_confirmation), "Confirmation not added");
+        ensure!(active_tx.confirmation.confirmations.contains(&new_confirmation), "Confirmation not added");
```
fluorostani, line 209: `confirmation.confirmations` ?

nahuseyoum: Confirmation is a property that contains confirmation related data such as the message hash and the list of confirmations. I could call it `confimration_data` but that was annoying me because its longer :)

PR #332, opened 2023-12-06
SYS-3773 move lower data to token manager and use eth-bridge as a service

nahuseyoum:
This PR moves out lower data logic into Token manager and only uses eth-bridge as a service to collect confirmations.


pallets/avn/src/lib.rs
```diff
@@ -677,6 +677,20 @@ pub trait BridgePublisher {
 
 pub trait OnBridgePublisherResult {
     fn process_result(tx_id: u32, caller_id: Vec<u8>, succeeded: bool) -> DispatchResult;
+    fn process_lower_proof_result(lower_id: u32, caller_id: Vec<u8>, abi_encoded_lower: Result<Vec<u8>, ()>) -> DispatchResult;
```
fluorostani, line 680: If we provide a default here then we wouldn't have to implement it as empty everywhere, right?

```suggestion
    fn process_lower_proof_result(lower_id: u32, caller_id: Vec<u8>, abi_encoded_lower: Result<Vec<u8>, ()>) -> DispatchResult {
    Ok(())
    }
```

PR #342, opened 2023-12-16
SYS-3773 Bug fix to handle bad transaction hashes and various improvements

nahuseyoum:
## Proposed changes

This PR fixes a bug where ethereum events gets stuck processing invalid transaction hashes. It also adds a new event to signal retries and automatically sets expiry period during a forkless upgrade.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [x] Documentation updated

### DIFF
--- a/pallets/eth-bridge/src/tx.rs
+++ b/pallets/eth-bridge/src/tx.rs
@@ -105,6 +105,12 @@ pub fn set_up_active_tx<T: Config>(req: SendRequestData) -> Result<(), Error<T>>
 }
 
 pub fn replay_send_request<T: Config>(mut tx: ActiveTransactionData<T>) -> Result<(), Error<T>> {
+    <crate::Pallet<T>>::deposit_event(Event::<T>::ActiveRequestRetried {
+        function_name: tx.request.function_name.clone(),
+        params: tx.request.params.clone(),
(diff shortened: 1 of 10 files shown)


ivan-cholakov (approved):

fluorostani (approved):

pallets/eth-bridge/src/tx.rs
```diff
@@ -105,6 +105,12 @@ pub fn set_up_active_tx<T: Config>(req: SendRequestData) -> Result<(), Error<T>>
 }
 
 pub fn replay_send_request<T: Config>(mut tx: ActiveTransactionData<T>) -> Result<(), Error<T>> {
```
alexpArtos, line 107: I see this is called from `finalize_state`, and that in turn is called from `add_corroboration`. But I don't see any limit to how many times we retry, or why we want to retry if we know the transaction has already failed.

nahuseyoum: If the sender is a malicious author, we retry until we find an honest author

PR #355, opened 2024-02-15
SYS-3828: ethereum events updated

ivan-cholakov:
updated pallet ethereum-events and changed how we handle EthEventId


thadouk (changes requested):
Why do we change EthEventId everywhere? I think we only need it changed on the genesis configuration block
`UncheckedEvents` will need a migration if we change the struct to a tupple.

pallets/ethereum-events/src/lib.rs
```diff
@@ -352,27 +351,27 @@ pub mod pallet {
...
-        pub event_challenge_period: T::BlockNumber,
+        pub event_challenge_period: BlockNumberFor<T>,
         pub lift_tx_hashes: Vec<H256>,
-        pub processed_events: Vec<(EthEventId, bool)>,
+        pub processed_events: Vec<((H256, H256), bool)>,
         pub nft_t1_contracts: Vec<(H160, ())>,
     }
 
-    #[cfg(feature = "std")]
+    // #[cfg(feature = "std")]
```
thadouk, line 360:
```suggestion
```

thadouk (approved):

PR #356, opened 2024-02-15
SYS-3828: updated pallet nft-manager

ivan-cholakov:
Updated nft-manager pallet


pallets/nft-manager/src/batch_nft.rs
```diff
@@ -136,7 +136,7 @@ pub fn encode_end_batch_sale_params<T: Config>(
 }
 
 pub fn process_mint_batch_nft_event<T: Config>(
-    event_id: &EthEventId,
+    event_id: &(H256, H256),
```
thadouk, line 139: SImilar feedback on using `EthEventId`
Its cleaner if we use the struct to know what hash is which.
The only issue with EthEventId is the additional encoding/decoding needed when its part of a genesis configuration.
The solution was to use a tuple instead in those usecases, maintaining the rest of the codebase the same.

pallets/nft-manager/src/lib.rs
```diff
@@ -128,15 +128,15 @@ pub mod pallet {
         pub _phantom: sp_std::marker::PhantomData<T>,
     }
 
-    #[cfg(feature = "std")]
+    // #[cfg(feature = "std")]
```
thadouk, line 131:
```suggestion
```

PR #359, opened 2024-02-15
SYS-3828: updated token manager

ivan-cholakov:
Updated token-manager


thadouk (commented):
The `EthEventId` changes need to be reverted here as well

PR #367, opened 2024-03-18
Sys 3925 fix primary validator

micaelffrancoAV:
This PR aims to fix the validator chosen to perform certain tasks. Before we used to have this based on block number and that would always return the same validator if the given block number was a multiple of 10.
This new implementation fixes this issue by always incrementing the validator everytime we call `calculate_primary_validator` with the aim of balancing the validator resources.

This PR includes:
- Fix to primary validator calculation (`calculate_primary_validator`).
- Fix to `isPrimary` function.
- Removed tests for the old logic and added new tests.

### DIFF
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -149,6 +149,12 @@ pub mod pallet {
         ErrorDecodingU32,
     }
 
+    #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeInfo)]
+    pub enum OperationType {
+        Ethereum,
+        Avn,
+    }
+
     #[pallet::storage]
     #[pallet::getter(fn validators)]
     /// The current set of validators (address and key) that may issue a transaction from the
@@ -163,20 +169,24 @@ pub mod pallet {
     #[pallet::getter(fn get_bridge_contract_address)]
     pub type AvnBridgeContractAddress<T: Config> = StorageValue<_, H160, ValueQuery>;
 
+    // #[pallet::storage]
+    // #[pallet::getter(fn get_primary_validator)]
+    // pub type PrimaryValidator<T: Config> = StorageValue<_, u8, ValueQuery>;
+
     #[pallet::storage]
     #[pallet::getter(fn get_primary_validator)]
-    pub type PrimaryValidator<T: Config> = StorageValue<_, u8, ValueQuery>;
+    pub type PrimaryValidator<T: Config> = StorageValue<_, (u8, u8), ValueQuery>;
 
     #[pallet::genesis_config]
     pub struct GenesisConfig<T: Config> {
         pub _phantom: sp_std::marker::PhantomData<T>,
         pub bridge_contract_address: H160,
-        pub primary_validator: u8,
+        pub primary_validator: (u8, u8),
     }
 
     impl<T: Config> Default for GenesisConfig<T> {
         fn default() -> Self {
-            Self { _phantom: Default::default(), bridge_contract_address: H160::zero(), primary_validator: 0 }
+            Self { _phantom: Default::default(), bridge_contract_address: H160::zero(), primary_validator: (0, 0) }
         }
     }
 
@@ -252,15 +262,22 @@ impl<T: Config> Pallet<T> {
 
     // TODO [TYPE: refactoring][PRI: LOW]: choose a better function name
     pub fn is_primary(
-        block_number: BlockNumberFor<T>,
+        op_type: OperationType,
         current_validator: &T::AccountId,
     ) -> Result<bool, Error<T>> {
-        let primary_validator = Self::calculate_primary_validator(block_number)?;
+        let validators = Self::validators();
+
+        let counters = PrimaryValidator::<T>::get();
+        let primary_validator = validators[match op_type {
+            OperationType::Ethereum => counters.0,
+            OperationType::Avn => counters.1,
+        } as usize].account_id.clone();
+
         return Ok(&primary_validator == current_validator)
     }
 
     pub fn calculate_primary_validator(
-        block_number: BlockNumberFor<T>,
+        op_type: OperationType
     ) -> Result<T::AccountId, Error<T>> {
         let validators = Self::validators();
 
@@ -269,10 +286,21 @@ impl<T: Config> Pallet<T> {
             return Err(Error::<T>::NoValidatorsFound)
         }
 
-        let index = PrimaryValidator::<T>::get();
-        PrimaryValidator::<T>::put((index + 1) % validators.len() as u8);
+        let mut counters = PrimaryValidator::<T>::get();
+        let validators_len = Self::validators().len() as u8;
 
-        return Ok(validators[index as usize].account_id.clone())
+        return match op_type {
+            OperationType::Ethereum => { // ethereum
+                counters.0 = (counters.0 + 1) % validators_len;
+                PrimaryValidator::<T>::put(&counters);
+                Ok(validators[counters.0 as usize].account_id.clone())
+            }
+            OperationType::Avn => { // avn
+                counters.1 = (counters.1 + 1) % validators_len;
+                PrimaryValidator::<T>::put(&counters);
+                Ok(validators[counters.1 as usize].account_id.clone())
+            }
+        };
     }
 
     pub fn get_validator_for_current_node() -> Option<Validator<T::AuthorityId, T::AccountId>> {
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -21,11 +21,19 @@ use alloc::{
     string::{String, ToString},
 };
 
-use frame_support::{dispatch::DispatchResult, log::*, traits::OneSessionHandler};
+use codec::{Decode, Encode};
+use core::convert::TryInto;
+use frame_support::{
+    dispatch::DispatchResult,
+    log::*,
+    pallet_prelude::{MaxEncodedLen, TypeInfo},
+    traits::OneSessionHandler,
+};
 use frame_system::{
     ensure_root,
     pallet_prelude::{BlockNumberFor, OriginFor},
 };
+pub use pallet::*;
 use sp_application_crypto::RuntimeAppPublic;
 use sp_avn_common::{
     bounds::MaximumValidatorsBound,
@@ -46,10 +54,6 @@ use sp_runtime::{
     DispatchError, WeakBoundedVec,
 };
 use sp_std::prelude::*;
-
-use codec::{Decode, Encode};
-use core::convert::TryInto;
-pub use pallet::*;
 
 #[path = "tests/testing.rs"]
 pub mod testing;
@@ -83,6 +87,12 @@ pub const MAX_VALIDATOR_ACCOUNTS: u32 = 10;
 
 pub const PACKED_LOWER_PARAM_SIZE: usize = 76;
 pub type LowerParams = [u8; PACKED_LOWER_PARAM_SIZE];
+
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, TypeInfo, MaxEncodedLen, Copy)]
+pub struct PrimaryValidatorData {
+    pub ethereum: u8,
+    pub avn: u8,
+}
 
 #[frame_support::pallet]
 pub mod pallet {
@@ -171,7 +181,7 @@ pub mod pallet {
 
     #[pallet::storage]
     #[pallet::getter(fn get_primary_validator)]
-    pub type PrimaryValidator<T: Config> = StorageValue<_, (u8, u8), ValueQuery>;
+    pub type PrimaryValidator<T: Config> = StorageValue<_, PrimaryValidatorData, ValueQuery>;
 
     #[pallet::genesis_config]
     pub struct GenesisConfig<T: Config> {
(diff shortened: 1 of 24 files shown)


pallets/avn/src/tests/tests.rs
```diff
@@ -1,64 +1,91 @@
...
+fn the_first_validator_is_picked_again_after_the_last_validator_for_ethereum_operations() {
     let mut ext = ExtBuilder::build_default().with_validators().as_externality();
     ext.execute_with(|| {
-        let block_number = 2;
-        let expected_primary = 3;
-        let result = AVN::is_primary(block_number, &expected_primary);
-        assert!(result.is_ok(), "Getting primary validator failed");
-        assert_eq!(result.unwrap(), true);
+        let validators = AVN::validators();
+        let eth_index_before: u8 = AVN::get_primary_validator().0;
```
alexpArtos, line 20: I'd please ask for this to be replaced by a single function somewhere, in all the tests, that retrieves the right index. For AVN and for Ethereum.
This is too low level, and if the actual implementation changes, there is a lot in the tests that will have to change as well. Even better, can we use a function from the actual code, if it already exists?

pallets/avn/src/tests/tests.rs
```diff
@@ -1,64 +1,91 @@
...
+            AVN::calculate_primary_validator(OperationType::Avn).unwrap();
+        }
+
+        let eth_index_after: u8 = AVN::get_primary_validator().1;
+        assert_eq!(eth_index_before, eth_index_after);
     });
 }
 
+#[test]
+fn is_primary_function_correctly_returns_the_current_primary_validator_for_avn_operations() {
```
alexpArtos, line 75: I don't understand the purpose of this test. What does the `is_primary_validator` do?
Shouldn't this function be returning true or false, and the tests check for that?
Also, `is_primary_validator` should **not** move the value of the next primary validator, since it should be a simple check, and not an action.
It would be valuable to have that test.

pallets/avn/src/tests/tests.rs
```diff
@@ -1,64 +1,91 @@
-use crate::mock::*;
+use crate::{mock::*, OperationType};
```
alexpArtos, line 1: Overall, I think these tests (even though they're clearly unit tests) could be written in a more BDD style:

- when `calculate_primary_validator` is called for operation Ethereum
  - the next validator for Ethereum increments by one, if the current one is not the last
  - the validator for Ethereum wraps around, if the current one is the last
  - the validator for AvN stays the same

- calling `is_primary_validator` for Ethereum
  - does not change the next_primary validator for Ethereum
  - returns true if the argument is the same as the next Ethereum validator
  - returns false if it is not

(repeat for the other operation)

pallets/eth-bridge/src/tx.rs
```diff
@@ -82,16 +83,17 @@ pub fn set_up_active_tx<T: Config>(req: SendRequestData) -> Result<(), Error<T>>
...
         confirmation: ActiveConfirmation { msg_hash, confirmations: BoundedVec::default() },
         tx_data: Some(ActiveEthTransaction {
-            function_name: req.function_name,
-            eth_tx_params: extended_params,
+            function_name: req.function_name.clone(),
+            eth_tx_params: extended_params.clone(),
             expiry,
             eth_tx_hash: H256::zero(),
-            sender: assign_sender()?,
+            sender: new_sender.clone(),
```
alexpArtos, line 96: why is this no longer inlined? I don't mind, but would like to know if there was a reason behind it

alexpArtos (changes requested):

pallets/eth-bridge/src/tx.rs
micaelffrancoAV: So we can use it on the event.

pallets/avn/src/lib.rs
```diff
@@ -163,22 +169,32 @@ pub mod pallet {
     #[pallet::getter(fn get_bridge_contract_address)]
     pub type AvnBridgeContractAddress<T: Config> = StorageValue<_, H160, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn get_primary_validator)]
```
nahuseyoum, line 173:
```suggestion
    #[pallet::getter(fn get_primary_collator)]
```

pallets/eth-bridge/src/lib.rs
```diff
@@ -195,6 +195,12 @@ pub mod pallet {
                 BoundedVec<(BoundedVec<u8, TypeLimit>, BoundedVec<u8, ValueLimit>), ParamsLimit>,
             caller_id: BoundedVec<u8, CallerIdLimit>,
         },
+        SetUpActiveRequest {
```
nahuseyoum, line 198: why is this needed here?

pallets/eth-bridge/src/tx.rs
```diff
@@ -101,6 +103,12 @@ pub fn set_up_active_tx<T: Config>(req: SendRequestData) -> Result<(), Error<T>>
         last_updated: <frame_system::Pallet<T>>::block_number(),
     });
 
+    <crate::Pallet<T>>::deposit_event(Event::<T>::SetUpActiveRequest {
```
nahuseyoum, line 106: Do we want to add an event here?

pallets/eth-bridge/src/lib.rs
micaelffrancoAV: I added this as a way to verify which validator was responsible for the request. Thought it might be useful but can remove it if not needed.

pallets/avn/src/lib.rs
```diff
@@ -246,28 +275,55 @@ impl<T: Config> Pallet<T> {
...
+        // If there are no validators there's no point continuing
+        if validators.len() == 0 {
+            return Err(Error::<T>::NoValidatorsFound)
+        }
+
+        let counters = PrimaryValidator::<T>::get();
+        let primary_validator = &validators[match op_type {
+            OperationType::Ethereum => counters.ethereum,
+            OperationType::Avn => counters.avn,
+        } as usize]
```
thadouk, line 292: We shouldn't trust the counters value blindly to access in memory values, cause it can lead to a panic.

the counters value get calculated via a mod operation with `validators_len`, but that value is not constant across time.
In a system with 5 collators, should a collator leave the system, and the storage item has him as the next one, the above line will cause a panic.

pallets/avn/src/tests/tests.rs
```diff
@@ -1,62 +1,242 @@
-use crate::mock::*;
+use crate::{
+    mock::{TestRuntime, *},
+    Error, OperationType,
+};
+use frame_support::{
+    assert_err,
+    assert_noop,
+    // dispatch::{DispatchError, DispatchResult},
```
thadouk, line 8:
```suggestion
```

thadouk (changes requested):

pallets/avn/src/lib.rs
micaelffrancoAV: Added a check to verify if the index is not out of bounds.

pallets/eth-bridge/src/tx.rs
micaelffrancoAV: I can remove it.

pallets/avn/src/lib.rs
```diff
@@ -246,28 +273,67 @@ impl<T: Config> Pallet<T> {
...
+        };
+
+        if index >= validators.len() {
+            return Err(Error::<T>::InvalidValidatorIndex)
+        };
+
+        let primary_validator = &validators[match op_type {
+            OperationType::Ethereum => counters.ethereum,
+            OperationType::Avn => counters.avn,
+        } as usize]
```
nahuseyoum, line 300:
```suggestion
        let primary_validator = &validators[index]
```

pallets/avn/src/lib.rs
```diff
@@ -246,28 +273,67 @@ impl<T: Config> Pallet<T> {
...
+
+        let counters = PrimaryValidator::<T>::get();
+
+        let index = match op_type {
+            OperationType::Ethereum => counters.ethereum as usize,
+            OperationType::Avn => counters.avn as usize,
+        };
+
+        if index >= validators.len() {
+            return Err(Error::<T>::InvalidValidatorIndex)
```
nahuseyoum, line 294: **this is an open question**: Instead of throwing an error here, can we reset the index to 0 and start again? Will this cause any issues? I know we should never get in this state but if we can self heal in a safe way it would be good.

pallets/eth-bridge/src/tx.rs
```diff
@@ -82,13 +83,14 @@ pub fn set_up_active_tx<T: Config>(req: SendRequestData) -> Result<(), Error<T>>
...
+    let new_sender = assign_sender()?;
 
     ActiveRequest::<T>::put(ActiveRequestData {
         request: Request::Send(req.clone()),
         confirmation: ActiveConfirmation { msg_hash, confirmations: BoundedVec::default() },
         tx_data: Some(ActiveEthTransaction {
-            function_name: req.function_name,
-            eth_tx_params: extended_params,
+            function_name: req.function_name.clone(),
+            eth_tx_params: extended_params.clone(),
```
nahuseyoum, line 93: do we need to clone here?

pallets/summary/src/challenge.rs
```diff
@@ -134,8 +135,9 @@ fn can_challenge<T: Config>(
         return false
     }
 
-    let is_chosen_validator = AVN::<T>::is_primary(ocw_block_number, &this_validator.account_id)
-        .unwrap_or_else(|_| false);
+    let is_chosen_validator =
+        AVN::<T>::is_primary(OperationType::Ethereum, &this_validator.account_id)
```
nahuseyoum, line 139: Should this be `AvN`? I can see in /summary/lib.rs we use `AvN` when computing the primary. Is that for a different purpose?

nahuseyoum (commented):
:+1:

pallets/eth-bridge/src/tx.rs
micaelffrancoAV: No :)

pallets/summary/src/challenge.rs
micaelffrancoAV: Should be Avn, yes.

pallets/avn/src/lib.rs
micaelffrancoAV: I think this makes sense. 
We should not have any issues by resetting the index, going back to zero in the end of the array would be the correct fix for this error because it will follow the modular approach we currently have when advancing index.

Thanks for the comment, I think we should apply it.

nahuseyoum (approved):

ivan-cholakov (approved):

pallets/avn/src/lib.rs
```diff
@@ -246,28 +273,62 @@ impl<T: Config> Pallet<T> {
...
-            .map_err(|_| Error::<T>::ErrorConvertingBlockNumber)?;
-
-        let index = block_number % validators.len();
-        return Ok(validators[index].account_id.clone())
+        let mut counters = PrimaryCollator::<T>::get();
+        let validators_len = Self::validators().len() as u8;
+
+        let index = match op_type {
+            OperationType::Ethereum => {
+                counters.ethereum = (counters.ethereum + 1) % validators_len;
```
thadouk, line 315:
```suggestion
                counters.ethereum = counters.ethereum.saturating_add(1) % validators_len;
```

pallets/avn/src/lib.rs
```diff
@@ -246,28 +273,62 @@ impl<T: Config> Pallet<T> {
...
+                PrimaryCollator::<T>::put(PrimaryCollatorData {
+                    ethereum: counters.ethereum,
+                    avn: counters.avn,
+                });
+                counters.ethereum
+            },
+            OperationType::Avn => {
+                counters.avn = (counters.avn + 1) % validators_len;
```
thadouk, line 323: can use saturating here as well.

thadouk (approved):

pallets/avn/src/lib.rs
```diff
@@ -246,28 +272,68 @@ impl<T: Config> Pallet<T> {
...
+        }
+
+        let mut counters = PrimaryCollator::<T>::get();
+
+        let mut index = match op_type {
+            OperationType::Ethereum => counters.ethereum as usize,
+            OperationType::Avn => counters.avn as usize,
+        };
+
+        if index >= validators.len() {
```
alexpArtos, line 292: It's probably ok to have this check here, but this is an indication that we did not do our work correctly somewhere else.
This can only happen if the validators have changed (one has been removed) after we've last chosen the primary collator. But then, that means that one action left the system in a broken state, which is not good.

How to solve this?
1. ensure that all actions that are potentially destructive clean up after them. Removing validators has to update the primary collator.

Analysis: I don't like this, because it means we have to know, at each specific site (removing validators) all the other parts of the logic that depend on it. It is not the competence of the validators manager to know that, and so it should not be this doing the update of primary collator.

2. make the operations associated with primary collator resilient to changes in their parameters:

This should be preferable. Ideally, we would not have to this check to update the primary collator in all the functions that make use of that storage item. Currently, we have this one (is_primary) but we semantically doing this verification and update here feels wrong.
This function should only check if a validator **is** the primary validator. But if there is ambiguity about what that primary is, because the value in storage may be inconsistent, then I think we should have instead a facade for that storage item, a function whose purpose it is to ensure it always reports a valid value.

This whole logic of checking could be migrated there. From the point of view of meaning of the code, it would be more correct. But we would have to discipline ourselves NOT to use the storage item directly.
From a performance aspect, every time we wanted to access the storage item we would do this check and update the item if needed. Most of times, we will do a couple of ifs and return the item's value.

The alternative, here, also incurs this extra cost, but every time we call `is_primary`. So depending on how often we call each, the cost may be the same or not. But it never goes away. So I'd go with what makes more logical sense.

alexpArtos (approved):

pallets/avn/src/lib.rs
alexpArtos: We should not need to obsess about saturations, should we? Why this push for using it everywhere?
The `%` ensures the numbers will never be larger than `validators_len`. Realistically, for us that is as small as it gets (`u8`). Why do we need to obsess about saturation at all?

alexpArtos (approved):

PR #368, opened 2024-03-19
Sys 3931 identify events

ivan-cholakov:
Added a function to query the passed smart contracts and get the required events based on the passed event signature


node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,61 @@
...
+    let web3 = web3::Web3::new(web3::transports::Http::new(eth_node_url).unwrap());
+
+    let filter = FilterBuilder::default()
+        .address(contract_addresses)
+        .topics(Some(event_signatures), None, None, None)
+        .from_block(web3::types::BlockNumber::Number(U64::from(start_block)))
+        .to_block(web3::types::BlockNumber::Number(U64::from(end_block)))
+        .build();
+
+    let logs = web3.eth().logs(filter).await.unwrap();
```
thadouk, line 26: This unwrap seems unsafe, can we make it return an error if result is not ok?

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,61 @@
...
+            match LiftedData::parse_bytes(Some(data.0), topics) {
+                Ok(data) => EventData::LogLifted(data),
+                Err(_) => return Err(Error::LiftedEventBadDataLength),
+            }
+        },
+    };
+
+    Ok(DiscoveredEvent {
+        event: EthEvent { event_id, event_data },
+        block: log.block_number.unwrap().as_u64(),
```
thadouk, line 59: same comment for unwrap.

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,61 @@
...
+        let discovered_event = parse_log(log)?;
+        events.push(discovered_event);
+    }
+
+    Ok(events)
+}
+
+fn parse_log(log: Log) -> Result<DiscoveredEvent, Error> {
+    let signature = log.topics[0];
+    let transaction_hash = log.transaction_hash.unwrap();
```
thadouk, line 40: same comment for unwrap

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+pub async fn identify_events(
+    eth_node_url: &str,
+    start_block: u64,
+    end_block: u64,
+    contract_addresses: Vec<H160>,
+    event_signatures: Vec<H256>,
+) -> Result<Vec<DiscoveredEvent>, Error> {
+    let web3 = web3::Web3::new(web3::transports::Http::new(eth_node_url).unwrap());
```
thadouk, line 17: can we make the web3 connection a parameter so we can reuse the existing one?

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    let event_data = match log.data {
+        data => {
+            let topics: Vec<Vec<u8>> = log.topics.iter().map(|t| t.0.to_vec()).collect();
+            match LiftedData::parse_bytes(Some(data.0), topics) {
+                Ok(data) => EventData::LogLifted(data),
+                Err(_) => return Err(Error::LiftedEventBadDataLength),
+            }
+        },
+    };
```
thadouk, line 55: I think the parsing should do all the logs cases, not just lifts.
Since we have already filtered the ones we want from the signature, then we should expect only the ones we had selected.

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,74 @@
...
+        .address(contract_addresses)
+        .topics(Some(event_signatures), None, None, None)
+        .from_block(web3::types::BlockNumber::Number(U64::from(start_block)))
+        .to_block(web3::types::BlockNumber::Number(U64::from(end_block)))
+        .build();
+
+        let logs_result = web3.eth().logs(filter).await;
+        let logs = match logs_result {
+            Ok(logs) => logs,
+            Err(err) => return Err(Error::NftTransferToEventWrongTopicCount)
```
nahuseyoum, line 28: The error we are returning doesn't seem to be correct.

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,74 @@
...
+        Some(transaction_hash) => transaction_hash,
+        None => return Err(Error::MissingTransactionHash),
+    };
+    
+    let event_id = EthEventId {
+        signature: sp_core::H256::from(signature.0),
+        transaction_hash: sp_core::H256::from(transaction_hash.0),
+    };
+
+    let event_data = match log.data {
```
nahuseyoum, line 55: what are we matching here?

node/avn-service/src/identify_eth_events.rs
nahuseyoum: I think the outer match should be on the `signature` and the inner match should be a `switch`?

node/avn-service/src/lib.rs
```diff
@@ -21,6 +21,7 @@ use tide::{http::StatusCode, Error as TideError};
 pub use web3Secp256k1::SecretKey as web3SecretKey;
 
 pub mod extrinsic_utils;
+pub mod identify_eth_events;
```
nahuseyoum, line 24: perhaps a better name would be `EthereumEventsHandler`?

node/avn-service/src/identify_eth_events.rs
```diff
@@ -0,0 +1,74 @@
...
+        data => {
+            let topics: Vec<Vec<u8>> = log.topics.iter().map(|t| t.0.to_vec()).collect();
+            match LiftedData::parse_bytes(Some(data.0), topics) {
```
nahuseyoum, line 58: This might be a good opportunity to parse the events natively using web3?

primitives/avn-common/src/event_types.rs
```diff
@@ -16,6 +16,12 @@ const BYTE_LENGTH: usize = 1; // the length of 1 byte
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum Error {
+    MissingBlockNumber,
+    MissingLogData,
```
thadouk, line 20:
```suggestion
```

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -0,0 +1,129 @@
+use sp_avn_common::event_types::{
+    AddedValidatorData, AvtGrowthLiftedData, Error, EthEvent, EthEventId, EventData, LiftedData,
+    NftCancelListingData, NftEndBatchListingData, NftMintData, NftTransferToData, ValidEvents,
+};
```
thadouk, line 4: regarding sp_avn_common::Error and expanding it with these additional use cases.
I think it would be cleaner, if we had here its own Error module and map any errors into equivalent ones here.
For example all parsing errors can be ParsingError.
It would also be beneficial if we had some error/info/trace logs here to help with future investigations, especially with the interactions with Ethereum as this has been a pain point in the past

thadouk (approved):

thadouk (approved):

PR #371, opened 2024-03-25
SYS-3930: Implement Vote Processing

thadouk:
## Proposed changes

This PR introduces support for processing votes upon achieving consensus. It extends extrinsics and data structures accordingly.

Once the final vote is cast, the voting round for the Ethereum range partition is finalized, and the partition is processed.

Additionally, this PR extends the subscription mechanism of the eth-bridge pallet. By default, pallets ignore processed events, requiring the implementation of a handler to process them.
Benchmarking for certain actions will be addressed in a separate PR.

Jira Tasks:
- SYS-3930
- SYS-3635

### DIFF
--- a/pallets/eth-bridge/src/lib.rs
+++ b/pallets/eth-bridge/src/lib.rs
@@ -138,7 +138,7 @@ pub mod pallet {
         traits::UnixTime,
         Blake2_128Concat,
     };
-    use sp_avn_common::event_types::{EthEvent, EthEventId};
+    use sp_avn_common::event_types::{EthEvent, EthEventId, ValidEvents};
 
     #[pallet::config]
     pub trait Config:
@@ -238,12 +238,10 @@ pub mod pallet {
     pub type ActiveEthereumRange<T: Config> = StorageValue<_, ActiveEthRange, OptionQuery>;
 
     #[pallet::storage]
-    pub type DiscoveredEventsFractions<T: Config> = StorageDoubleMap<
+    pub type EthereumEvents<T: Config> = StorageMap<
         _,
         Blake2_128Concat,
-        EthBlockRange,
-        Blake2_128Concat,
-        DiscoveredEthEventsFraction,
+        EthereumEventsPartition,
         BoundedBTreeSet<T::AccountId, VotesLimit>,
         ValueQuery,
     >;
@@ -310,7 +308,6 @@ pub mod pallet {
         EventVotesFull,
         InvalidEventVote,
         EventVoteExists,
-        EventScanningDisabled,
         NonActiveEthereumRange,
         VotingEnded,
     }
@@ -517,36 +514,39 @@ pub mod pallet {
         pub fn submit_ethereum_events(
             origin: OriginFor<T>,
             author: Author<T>,
-            range: EthBlockRange,
-            events_fraction: DiscoveredEthEventsFraction,
+            events_partition: EthereumEventsPartition,
             _signature: <T::AuthorityId as RuntimeAppPublic>::Signature,
         ) -> DispatchResultWithPostInfo {
             ensure_none(origin)?;
 
-            ensure!(events_fraction.is_valid(), Error::<T>::InvalidEventVote);
-            // Validate that the range is the current one.
-            let active_range =
-                Self::active_ethereum_range().ok_or(Error::<T>::EventScanningDisabled)?;
-            ensure!(range == active_range.range, Error::<T>::NonActiveEthereumRange);
-            ensure!(
-                active_range.status == EventProcessingStatus::UnderValidation,
-                Error::<T>::VotingEnded
-            );
+            let active_range = match Self::active_ethereum_range() {
+                Some(active_range) => {
+                    ensure!(
+                        *events_partition.range() == active_range.range &&
+                            events_partition.partition() == active_range.partition,
+                        Error::<T>::NonActiveEthereumRange
+                    );
+                    active_range
+                },
+                None => Default::default(),
+            };
 
-            // Check if already vote has been casted
-            let votes = DiscoveredEventsFractions::<T>::get(&range, &events_fraction);
+            if active_range.is_initial_range() == false {
+                ensure!(
+                    has_author_casted_event_vote::<T>(&author.account_id) == false,
+                    Error::<T>::EventVoteExists
+                );
+            }
 
-            ensure!(votes.contains(&author.account_id) == false, Error::<T>::EventVoteExists);
-            let quorum = AVN::<T>::quorum() as usize;
-            ensure!(votes.len() <= quorum, Error::<T>::EventVotesFull);
+            let mut votes = EthereumEvents::<T>::get(&events_partition);
+            votes.try_insert(author.account_id).map_err(|_| Error::<T>::EventVotesFull)?;
 
-            DiscoveredEventsFractions::<T>::try_mutate(&range, &events_fraction, |votes| {
-                votes.try_insert(author.account_id)
-            })
-            .map_err(|_| Error::<T>::EventVotesFull)?;
-
-            // Process results ensure all partitions are accepted
-            complete_voting_round_if_ready::<T>(range, quorum, events_fraction.id());
+            if votes.len() < AVN::<T>::quorum() as usize {
+                EthereumEvents::<T>::insert(&events_partition, votes);
+            } else {
+                process_ethereum_events_partition::<T>(&active_range, &events_partition);
+                advance_partition::<T>(&active_range, &events_partition);
+            }
 
             Ok(().into())
         }
@@ -560,46 +560,6 @@ pub mod pallet {
                     log::error!("❌ Error processing currently active request: {:?}", e);
                 }
             }
-        }
-
-        fn on_idle(_block: BlockNumberFor<T>, mut remaining_weight: Weight) -> Weight {
-            if remaining_weight.checked_reduce(T::DbWeight::get().reads(1)).is_none() {
-                return remaining_weight
-            }
-            remaining_weight.saturating_reduce(T::DbWeight::get().reads(1));
-            let active_range = match Self::active_ethereum_range() {
-                Some(range) => range,
-                None => return remaining_weight,
-            };
-
-            // Ensure enough capacity is there for doing the processing.
-            let partitions_count = match active_range.status {
-                EventProcessingStatus::Accepted(partitions_count) => partitions_count,
-                _ => return remaining_weight,
-            };
-
-            // TODO replace zero with benchmark value
-            let process_weight =
-                T::DbWeight::get().reads(partitions_count as u64).saturating_add(Weight::zero());
-            if remaining_weight.checked_reduce(process_weight).is_none() {
-                return remaining_weight
-            }
-
-            // Read all, then sort, then start processing one by one.
-            let mut partitions =
-                DiscoveredEventsFractions::<T>::iter_key_prefix(&active_range.range)
-                    .collect::<Vec<DiscoveredEthEventsFraction>>();
-            partitions.sort_by(|a, b| a.fraction().cmp(&b.fraction()));
-            remaining_weight.saturating_reduce(process_weight);
-
-            let next_partition = match partitions.first() {
-                Some(partition) => partition,
-                None => return remaining_weight,
-            };
-
-            process_ethereum_events_fraction::<T>(&active_range.range, next_partition);
-
-            remaining_weight
         }
     }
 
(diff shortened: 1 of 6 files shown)


ivan-cholakov (approved):

pallets/eth-bridge/src/lib.rs
```diff
@@ -514,22 +523,30 @@ pub mod pallet {
...
             ensure_none(origin)?;
 
+            ensure!(events_fraction.is_valid(), Error::<T>::InvalidEventVote);
             // Validate that the range is the current one.
             let active_range =
                 Self::active_ethereum_range().ok_or(Error::<T>::EventScanningDisabled)?;
-            ensure!(range == active_range, Error::<T>::EventScanningDisabled);
+            ensure!(range == active_range.range, Error::<T>::NonActiveEthereumRange);
+            ensure!(
+                active_range.status == EventProcessingStatus::UnderValidation,
```
nahuseyoum, line 532:
```suggestion
                active_range.status == EventProcessingStatus::ValidationInProgress,
```
This might be a personal choice but when I read `UnderValidation` I thought it mean its below the required threshold :)

pallets/eth-bridge/src/lib.rs
```diff
@@ -514,22 +523,30 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             ensure_none(origin)?;
 
+            ensure!(events_fraction.is_valid(), Error::<T>::InvalidEventVote);
```
nahuseyoum, line 526: do we want to specify that the events we got here are a "fraction" of the total? Would it be simpler if we called it `ethereum_events`?

pallets/eth-bridge/src/lib.rs
```diff
@@ -514,22 +523,30 @@ pub mod pallet {
...
-            ensure!(range == active_range, Error::<T>::EventScanningDisabled);
+            ensure!(range == active_range.range, Error::<T>::NonActiveEthereumRange);
+            ensure!(
+                active_range.status == EventProcessingStatus::UnderValidation,
+                Error::<T>::VotingEnded
+            );
 
             // Check if already vote has been casted
-            let votes = DiscoveredEventsShard::<T>::get(&range, &events_fraction);
+            let votes = DiscoveredEventsFractions::<T>::get(&range, &events_fraction);
```
nahuseyoum, line 537: Personally I would rename `DiscoveredEventsFractions` to `EthereumEvents`

pallets/eth-bridge/src/lib.rs
```diff
@@ -514,22 +523,30 @@ pub mod pallet {
...
+
             ensure!(votes.contains(&author.account_id) == false, Error::<T>::EventVoteExists);
-            ensure!(votes.len() as u32 >= AVN::<T>::quorum(), Error::<T>::EventVotesFull);
+            let quorum = AVN::<T>::quorum() as usize;
+            ensure!(votes.len() <= quorum, Error::<T>::EventVotesFull);
```
nahuseyoum, line 541: We could be forgiving here and just do nothing instead of erroring?

pallets/eth-bridge/src/lib.rs
```diff
@@ -544,6 +561,46 @@ pub mod pallet {
                 }
             }
         }
+
+        fn on_idle(_block: BlockNumberFor<T>, mut remaining_weight: Weight) -> Weight {
```
nahuseyoum, line 565: I am not sure if we want to de-prioritise Ethereum events by using onIdle. IMO we want to process them as quickly as possible.

primitives/avn-common/src/event_discovery.rs
```diff
@@ -54,33 +68,43 @@ impl Ord for DiscoveredEvent {
     }
 }
 
+pub type FractionsCount = u16;
```
nahuseyoum, line 71: does it make sense to have a named type for this primitive value?

primitives/avn-common/src/event_discovery.rs
```diff
@@ -54,33 +68,43 @@ impl Ord for DiscoveredEvent {
...
 type EthEventsPartition = BoundedBTreeSet<DiscoveredEvent, EventsBatchLimit>;
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
 pub struct DiscoveredEthEventsFraction {
-    data: EthEventsPartition,
-    fraction: u16,
-    fraction_count: u16,
     id: H256,
+    fraction: FractionsCount,
```
nahuseyoum, line 76: IMO this makes it hard to read the code when compared to `fraction: u16`

primitives/avn-common/src/event_discovery.rs
```diff
@@ -54,33 +68,43 @@ impl Ord for DiscoveredEvent {
...
+    fraction_count: FractionsCount,
```
nahuseyoum, line 77:
```suggestion
    max_number_of_events_per_report: FractionsCount,
```

primitives/avn-common/src/event_discovery.rs
```diff
@@ -14,6 +15,19 @@ pub struct EthBlockRange {
...
+        }
+    }
+    pub fn range(&self) -> (u32, u32) {
+        let end_block = self.start_block.saturating_add(self.length).saturating_less_one();
+        (self.start_block, end_block)
+    }
+}
+
 #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug, Default, TypeInfo, MaxEncodedLen)]
 pub struct DiscoveredEvent {
```
nahuseyoum, line 32:
```suggestion
pub struct EthereumEventsReport {
```

primitives/avn-common/src/event_discovery.rs
```diff
@@ -54,33 +68,43 @@ impl Ord for DiscoveredEvent {
...
 
 impl DiscoveredEthEventsFraction {
-    pub fn events(&self) -> &EthEventsPartition {
-        &self.data
+    pub fn id(&self) -> &H256 {
+        &self.id
     }
 
-    pub fn fraction(&self) -> u16 {
+    pub fn fraction(&self) -> FractionsCount {
```
nahuseyoum, line 86: whats the difference between `fraction` and `fraction_count`?

nahuseyoum (commented):
IMO I would remove the `chunking`/ `fraction` logic in the runtime. I think the complexity it adds is significant. Perhaps we can move this logic to the node and pick the `endBlock` of the range based on the reports we get from the collators. Happy to discuss this offline

pallets/eth-bridge/src/lib.rs
```diff
@@ -229,10 +235,10 @@ pub mod pallet {
 
     #[pallet::storage]
     #[pallet::getter(fn active_ethereum_range)]
-    pub type ActiveEthereumRange<T: Config> = StorageValue<_, EthBlockRange, OptionQuery>;
+    pub type ActiveEthereumRange<T: Config> = StorageValue<_, ActiveEthRange, OptionQuery>;
 
     #[pallet::storage]
-    pub type DiscoveredEventsShard<T: Config> = StorageDoubleMap<
+    pub type DiscoveredEventsFractions<T: Config> = StorageDoubleMap<
```
alexpArtos, line 241: what is an Events Fraction? Is this a processing batch?

pallets/eth-bridge/src/lib.rs
alexpArtos: yes, fraction has the sense of division, of quotient to me. A part of a whole that will not work until all fractions are brought together. But each event is valid on its own, isn't it? Perhaps this is just a collection a events, and there will be many of that type and size. I like `batch` for that, but lots of words would work: set, lot, group, etc.
You changed `shard` and maybe you had a reason for it. It is a loaded term in computer science, and may imply a number of meanings we don't want, so I'm ok with changing from it.

pallets/eth-bridge/src/lib.rs
alexpArtos: is `votes` the number of collators that have validated the fraction? What is the problem if we have more votes than needed?

pallets/eth-bridge/src/lib.rs
```diff
@@ -544,6 +561,46 @@ pub mod pallet {
...
+                T::DbWeight::get().reads(partitions_count as u64).saturating_add(Weight::zero());
+            if remaining_weight.checked_reduce(process_weight).is_none() {
+                return remaining_weight
+            }
+
+            // Read all, then sort, then start processing one by one.
+            let mut partitions =
+                DiscoveredEventsFractions::<T>::iter_key_prefix(&active_range.range)
+                    .collect::<Vec<DiscoveredEthEventsFraction>>();
+            partitions.sort_by(|a, b| a.fraction().cmp(&b.fraction()));
```
alexpArtos, line 592: Is this necessary? Or even desirable? We will be repeating part of the same sorting work every time we run this hook, I believe, if we don't process all of the partitions in the time we have available. Is there a strict need to process the events in any given order?

primitives/avn-common/src/event_discovery.rs
alexpArtos: I agree with Nahu. The meaning of 'fractions count' should be communicated by the variable name, not by the type. The type is more like 'units' of measurement, if you like, and I think it would only be relevant if we had lots of operations between several variables of the same 'type'

alexpArtos (approved):

pallets/eth-bridge/src/lib.rs
thadouk: >We could be forgiving here and just do nothing instead of erroring?
What is the benefit of being forgiving? We save space if we just reject it.
>is votes the number of collators that have validated the fraction? What is the problem if we have more votes than needed?
yes

pallets/eth-bridge/src/lib.rs
thadouk: We can't process all at once to avoid exceed block limits.
First fraction is processed at once. the rest are added on on_idle.

pallets/eth-bridge/src/lib.rs
thadouk: yes, its a requirement to process the events in the right order as some nft operations (transfer of ownership) has different result based on the order.

primitives/avn-common/src/event_discovery.rs
thadouk: Main reason I added this is this:
https://github.com/AventusProtocolFoundation/avn-parachain/pull/371/files#diff-b16d6c46e1465c34530cee63d8a1b6cafdf9cd5e55e481ef95c83202734895bbR132
It improves readability on the data struct.
I can remove though if that is the preference

primitives/avn-common/src/event_discovery.rs
thadouk: fraction_count is how many fractions has this broken into. Fraction is which of the fraction is it.
Like saying 1 out of 5 fractions

pallets/eth-bridge/src/lib.rs
thadouk: effectively yes.
From the total of events found, it is broken down into fractions of the whole, then voted and processed individually.

primitives/avn-common/src/event_discovery.rs
thadouk: I will say no to the suggestion. Its a single event that has been discovered and the block it was included.

primitives/avn-common/src/event_discovery.rs
thadouk: that is not what the fraction count is.
Its that total number of fractions

pallets/eth-bridge/src/lib.rs
thadouk: I think not. Given the definition of this variable:
```
events_fraction: DiscoveredEthEventsFraction,
```
Given its type, events_fraction is a clear name.

MBrozhko34 (approved):

pallets/eth-bridge/src/lib.rs
```diff
@@ -651,6 +668,81 @@ pub mod pallet {
 
         Ok(())
     }
+    fn has_author_casted_event_vote<T: Config>(author: &T::AccountId) -> bool {
```
alexpArtos, line 673: again, no vote identifier. Does any vote count? We are ignoring the partition differences themselves, so any vote anywhere will make this return true. What is the meaning of this function, overall?

primitives/avn-common/src/event_discovery.rs
```diff
@@ -90,29 +108,28 @@ pub mod events_helpers {
...
 
-        let _ = iter.try_for_each(|(fraction, chunk)| -> Result<(), ()> {
+        let _ = iter.try_for_each(|(partition, chunk)| -> Result<(), ()> {
             let inner_data: BTreeSet<DiscoveredEvent> = chunk.iter().cloned().collect();
             let data = EthEventsPartition::try_from(inner_data)?;
-            fractions.push(DiscoveredEthEventsFraction::new(
+            partitions.push(EthereumEventsPartition::new(
+                range.clone(),
+                partition as u16,
+                partitions_count == partition.saturating_add(1),
```
alexpArtos, line 128: does this `saturating_add` actually change the value of `partition` or not? This is relevant, because that would be introducing a side-effect (which is discouraged in functional programming, because in general it makes validating code correctness harder) that is beyond what we need in here.
I don't know exactly how saturating methods work, since they have an implicit argument (the self) and the look of them _could mean either implementation_, but the look of it suggests it changes the callee.
So, I'm not comfortable with using them unless we absolutely need to.

pallets/eth-bridge/src/lib.rs
```diff
@@ -508,28 +514,39 @@ pub mod pallet {
...
+                },
+                None => Default::default(),
+            };
 
-            // Check if already vote has been casted
-            let votes = DiscoveredEventsShard::<T>::get(&range, &events_fraction);
-            ensure!(votes.contains(&author.account_id) == false, Error::<T>::EventVoteExists);
-            ensure!(votes.len() as u32 >= AVN::<T>::quorum(), Error::<T>::EventVotesFull);
+            if active_range.is_initial_range() == false {
+                ensure!(
```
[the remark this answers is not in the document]
thadouk: yes.
Once the active range is initialised, meaning consensus has been achieved on runtime for a partition, then we can apply these constrains.
The first range will be calculated based on the latest head of tier1 for each collator in a deterministic way. Depending on availability of the node, and its access to tier1 there are edge cases where nodes might cast their initial vote in an earlier range. That is acceptable at this initialising phase.
Once a starting point is agreed through the authors, we can start enforcing the voting contrains.

primitives/avn-common/src/event_discovery.rs
```diff
@@ -56,32 +69,37 @@ impl Ord for DiscoveredEvent {
 
 type EthEventsPartition = BoundedBTreeSet<DiscoveredEvent, EventsBatchLimit>;
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
-pub struct DiscoveredEthEventsFraction {
+pub struct EthereumEventsPartition {
+    range: EthBlockRange,
+    partition: u16,
+    is_last: bool,
     data: EthEventsPartition,
```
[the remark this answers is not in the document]
thadouk: I can rename to `EthEventsSet` 
I would avoid inlining as it bloats the code.

primitives/avn-common/src/event_discovery.rs
thadouk: https://docs.rs/sp-runtime/latest/sp_runtime/traits/trait.Saturating.html#tymethod.saturating_add
self is immutable, no implementation can change `self` values.
See https://docs.rs/sp-runtime/latest/sp_runtime/traits/trait.Saturating.html#method.saturating_inc that does.
I can make the implementation use an immutable iterator to the partitions if that will give you more confidence.

pallets/eth-bridge/src/lib.rs
```diff
@@ -651,6 +670,81 @@ pub mod pallet {
 
         Ok(())
     }
+    fn has_author_casted_event_vote<T: Config>(author: &T::AccountId) -> bool {
+        for (_partition, votes) in EthereumEvents::<T>::iter() {
```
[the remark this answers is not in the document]
thadouk: Since each collator can vote once, we iterate through a small number of elements.
If we use the partition data, that would allow a collator to cast as many votes as he wants, bloating the storage.

PR #390, opened 2024-04-24
SYS-3965: added first stage migration of processed events mechanism to eth-bridge

MBrozhko34:
## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


pallets/eth-bridge/src/lib.rs
```diff
@@ -731,23 +733,33 @@ pub mod pallet {
         }
     }
 
-    fn process_ethereum_event<T: Config>(event: &EthEvent) {
+    fn process_ethereum_event<T: Config>(event: &EthEvent) -> Result<(), DispatchError> {
```
thadouk, line 736: i think we should consume any errors here, rather than bubbling up the error

pallets/token-manager/src/mock.rs
```diff
@@ -376,6 +377,10 @@ impl ProcessedEventsChecker for TestRuntime {
     fn check_event(event_id: &EthEventId) -> bool {
         return PROCESSED_EVENTS.with(|l| l.borrow_mut().iter().any(|event| event == event_id))
     }
+    
+    fn add_event(event_id: &EthEventId, processed: bool) -> DispatchResult {
```
thadouk, line 381: do we need an overwrite of the default one?

pallets/validators-manager/src/tests/mock.rs
```diff
@@ -350,6 +351,10 @@ impl ProcessedEventsChecker for TestRuntime {
             })
         })
     }
+    
+    fn add_event(event_id: &EthEventId, processed: bool) -> DispatchResult {
```
thadouk, line 355: ditto

pallets/token-manager/src/mock.rs
MBrozhko34: It is not possible to implement only one of the trait items without the other, this change was to resolve errors occurring as a result of adding trait functions

MBrozhko34:
@nahuseyoum I have written some comments for the tests. So far I had created the values in the mock (although partially incorrectly as the successful test case does not pass) and identified the edge cases. The tests themselves still need to reach the required function process_ethereum_event().

ivan-cholakov (approved):

pallets/eth-bridge/src/lib.rs
```diff
@@ -223,8 +223,11 @@ pub mod pallet {
                 BoundedVec<(BoundedVec<u8, TypeLimit>, BoundedVec<u8, ValueLimit>), ParamsLimit>,
             caller_id: BoundedVec<u8, CallerIdLimit>,
         },
-        /// EventProcessed(bool, EthEventId)
-        EventProcessed {
+        EventProcessingAccepted {
+            accepted: bool,
```
nahuseyoum, line 227: I think we can remove the boolean because the name suggest the outcome

pallets/eth-bridge/src/lib.rs
```diff
@@ -223,8 +223,11 @@ pub mod pallet {
                 BoundedVec<(BoundedVec<u8, TypeLimit>, BoundedVec<u8, ValueLimit>), ParamsLimit>,
             caller_id: BoundedVec<u8, CallerIdLimit>,
         },
-        /// EventProcessed(bool, EthEventId)
-        EventProcessed {
+        EventProcessingAccepted {
```
nahuseyoum, line 226: I would also re-use the same name as EthereumEvents: `EventAccepted` and `EventRejected`.

pallets/eth-bridge/src/lib.rs
```diff
@@ -824,28 +830,30 @@ pub mod pallet {
...
+        if T::ProcessedEventsChecker::processed_event_exists(&event.event_id.clone()) {
+            log::error!("💔 Event already processed, duplicate event");
 
-        <Pallet<T>>::deposit_event(Event::<T>::EventProcessed {
-            accepted: event_accepted,
-            eth_event_id: event.event_id.clone(),
-        });
-        // Add record of succesful processing via ProcessedEventsChecker
-        T::ProcessedEventsChecker::add_processed_event(&event.event_id.clone(), event_accepted);
+            <Pallet<T>>::deposit_event(Event::<T>::DuplicateEventSubmission {
```
nahuseyoum, line 839: I see why you are emitting an event here, I am assuming its because we don't want to error here which is correct. But in this particular case, the event we are processing has already been processed so Its "safe" to return an error and undo any storage changes related to processing this event.

pallets/eth-bridge/src/tests/mock.rs
```diff
@@ -181,8 +184,28 @@ pub fn create_confirming_author(author_id: u64) -> Author<TestRuntime> {
...
+        .unwrap();
+    EthereumEventsPartition::new(
+        EthBlockRange { start_block: 1, length: 1000 },
+        0,
+        false,
+        partition,
+    )
+}
+
+pub fn create_second_mock_event_partition(events: EthEvent) -> EthereumEventsPartition {
```
nahuseyoum, line 198: You can update `create_mock_event_partition` to take an extra argument. This avoids creating a second function.

nahuseyoum (approved):

pallets/avn/src/lib.rs
```diff
@@ -703,13 +703,17 @@ impl<ValidatorId: Member> Enforcer<ValidatorId> for () {
 }
 
 pub trait ProcessedEventsChecker {
-    fn check_event(event_id: &EthEventId) -> bool;
+    fn processed_event_exists(event_id: &EthEventId) -> bool;
+
```
thadouk, line 707:
```suggestion
```

pallets/eth-bridge/src/lib.rs
```diff
@@ -219,9 +223,13 @@ pub mod pallet {
...
-        /// EventProcessed(bool, EthEventId)
-        EventProcessed {
-            accepted: bool,
+        EventAccepted {
+            eth_event_id: EthEventId,
+        },
+        EventRejected {
+            eth_event_id: EthEventId,
+        },
+        DuplicateEventSubmission {
```
thadouk, line 232: this is not used.

pallets/eth-bridge/src/lib.rs
```diff
@@ -296,13 +304,16 @@ pub mod pallet {
     pub enum Error<T> {
         CorroborateCallFailed,
         DuplicateConfirmation,
+        DuplicateEventSubmission,
```
thadouk, line 307: this is not used

pallets/eth-bridge/src/lib.rs
```diff
@@ -219,9 +223,13 @@ pub mod pallet {
...
         },
-        /// EventProcessed(bool, EthEventId)
-        EventProcessed {
-            accepted: bool,
+        EventAccepted {
+            eth_event_id: EthEventId,
+        },
+        EventRejected {
+            eth_event_id: EthEventId,
+        },
```
thadouk, line 231: why use two events for signaling the processing events?

pallets/eth-bridge/src/lib.rs
```diff
@@ -794,48 +806,62 @@ pub mod pallet {
     fn process_ethereum_events_partition<T: Config>(
         active_range: &ActiveEthRange,
         partition: &EthereumEventsPartition,
-    ) {
+    ) -> Result<(), DispatchError> {
```
thadouk, line 809: This function should be infallible.
Once a voting round is concluded, the processing of each event should not affect the processing of the batch.

The errors should be handled, the appropriate events emited and processing continues.
Otherwise, the chain will be stuck in a loop of voting for the last vote and rejecting the processing.

pallets/eth-bridge/src/lib.rs
```diff
@@ -794,48 +806,62 @@ pub mod pallet {
...
         }
+
+        Ok(())
     }
 
-    fn process_ethereum_event<T: Config>(event: &EthEvent) {
-        // TODO before processing ensure that the event has not already been processed
-        match T::BridgeInterfaceNotification::on_event_processed(&event) {
+    fn process_ethereum_event<T: Config>(event: &EthEvent) -> Result<(), DispatchError> {
+        ensure!(
```
thadouk, line 839: we shouldn't error here. If the event has been processed, we avoid double processing, signal/log/emit the appropriate info and return.

thadouk (changes requested):

pallets/eth-bridge/src/lib.rs
MBrozhko34: Would it not be due to the fact that there are two possibilities in the event being accepted vs rejected and to distinguish between these two circumstances?

pallets/eth-bridge/src/lib.rs
nahuseyoum: If we call this function with an event that has already been processed, it's because the upstream functions did not validate properly.

pallets/eth-bridge/src/lib.rs
nahuseyoum: We have connected components that are currently looking for these events and its a lesser change to maintain the setup. Its also easier to search and filter instead of using event args.

pallets/eth-bridge/src/lib.rs
thadouk: if the authors/collators have reached consensus on the tier1 state then we shouldn't block it.
There are valid cases that this can occur:
1. events added and being processed while upgrading.
2. events added to tier2 due starting from an earlier last block
We should definitely mark the case, but not block the event processing mechanism.
The error will cause the last vote to be stuck and recasted, only to be rejected again and again in a loop.

nahuseyoum: In both of these scenarios, and possibly others, I would expect the collators to check first if the event is valid before submitting it (ex: already processed or `0x000`..). 
But in any case, this function is too low level to be required to be infallible. I think we can still throw an error here and catch it on `process_ethereum_events_partition` and log it there and move on to the next event? 
It feels like making an distinction between a bad event (one that is already processed, or is 0 
 - something we should add in the validation i guess :smile: ) and a valid event that fails to be processed makes sense which is what this low level function is doing.

MBrozhko34: I am happy with either solution, although perhaps handling the error within the `process_ethereum_events_partition` function is the best solution.

MBrozhko34: I have pushed a [commit](https://github.com/AventusProtocolFoundation/avn-parachain/pull/390/commits/4b74809faf44e3f2181cea8496b8180eab997bb4) the changes @nahuseyoum suggested @thadouk . This can be reverted if necessary although I believe it should be ok

pallets/eth-bridge/src/lib.rs
```diff
@@ -801,15 +813,25 @@ pub mod pallet {
...
-                        process_ethereum_event::<T>(&discovered_event.event);
+                        match process_ethereum_event::<T>(&discovered_event.event) {
+                            Ok(_) => {},
+                            Err(_) => {
+                                log::error!("💔 Duplicate Event Submission");
+                                <Pallet<T>>::deposit_event(Event::<T>::DuplicateEventSubmission {
+                                    eth_event_id: discovered_event.event.event_id.clone(),
+                                });
+                            },
+                        }
```
thadouk, line 824:
```suggestion
                        if process_ethereum_event::<T>(&discovered_event.event).is_err() {
                                log::error!("💔 Duplicate Event Submission");
                                <Pallet<T>>::deposit_event(Event::<T>::DuplicateEventSubmission {
                                    eth_event_id: discovered_event.event.event_id.clone(),
                                });
                        }
```

thadouk (approved):

pallets/eth-bridge/src/lib.rs
nahuseyoum: Maybe we should have a generic `InvalidEventSubmission` instead of `DuplicateEventSubmission`? We can then use this for 0 transaction hashes too (not just existing events)

PR #391, opened 2024-04-25
Introduces initial range consesnus mechanism

thadouk:
## Proposed changes

Adds a new endpoint that allows the chain to reach consenus over the initial ethereum range to be used.
This allows to simplify the voting logic for active ranges, eliminating the special case of the first range.

### DIFF
--- a/pallets/eth-bridge/src/lib.rs
+++ b/pallets/eth-bridge/src/lib.rs
@@ -251,6 +251,15 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    #[pallet::storage]
+    pub type SubmittedBlockNumbers<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        EthBlockRange,
+        BoundedBTreeSet<T::AccountId, VotesLimit>,
+        ValueQuery,
+    >;
+
     #[pallet::genesis_config]
     pub struct GenesisConfig<T: Config> {
         pub _phantom: sp_std::marker::PhantomData<T>,
@@ -563,7 +572,7 @@ pub mod pallet {
             ensure_none(origin)?;
             ensure!(Self::active_ethereum_range().is_none(), Error::<T>::VotingEnded);
             ensure!(
-                author_has_cast_event_vote::<T>(&author.account_id) == false,
+                author_has_submitted_latest_block::<T>(&author.account_id) == false,
                 Error::<T>::EventVoteExists
             );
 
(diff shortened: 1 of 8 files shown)


pallets/eth-bridge/src/lib.rs
```diff
@@ -556,6 +549,69 @@ pub mod pallet {
 
             Ok(().into())
         }
+
+        #[pallet::call_index(7)]
+        #[pallet::weight(<T as Config>::WeightInfo::add_eth_tx_hash())]
```
nahuseyoum, line 554:
```suggestion
         // TODO: Benchmark me
        #[pallet::weight(0)]
```

pallets/eth-bridge/src/lib.rs
```diff
@@ -884,26 +943,26 @@ impl<T: Config> Pallet<T> {
                     active_range.event_types_filter.into_iter().collect::<Vec<ValidEvents>>();
 
                 let decoded_hex =
-                hex::decode("418da8f85cfa851601f87634c6950491b6b8785a6445c8584f5658048d512cae").
-                expect("test"); 
+                    hex::decode("418da8f85cfa851601f87634c6950491b6b8785a6445c8584f5658048d512cae")
```
nahuseyoum, line 946: I know this is existing code, but what does this represent? Shouldn't we be handing any errors correctly or emitting a more descriptive `expect` message?

pallets/eth-bridge/src/lib.rs
```diff
@@ -556,6 +549,69 @@ pub mod pallet {
...
+                author_has_cast_event_vote::<T>(&author.account_id) == false,
+                Error::<T>::EventVoteExists
+            );
+
+            let nominated_range =
+                events_helpers::compute_finalised_block_range_for_latest_ethereum_block(
+                    latest_seen_block,
+                );
+            if let Some(events_partition) =
+                events_helpers::discovered_eth_events_partition_factory(nominated_range, Vec::new())
```
nahuseyoum, line 573: Why are we calling this function with an empty array?

pallets/eth-bridge/src/lib.rs
```diff
@@ -556,6 +549,69 @@ pub mod pallet {
...
+        #[pallet::call_index(7)]
+        #[pallet::weight(<T as Config>::WeightInfo::add_eth_tx_hash())]
+        pub fn submit_latest_ethereum_block(
+            origin: OriginFor<T>,
+            author: Author<T>,
+            latest_seen_block: u32,
+            _signature: <T::AuthorityId as RuntimeAppPublic>::Signature,
+        ) -> DispatchResultWithPostInfo {
+            ensure_none(origin)?;
+            ensure!(Self::active_ethereum_range().is_some(), Error::<T>::VotingEnded);
```
nahuseyoum, line 562: Should this be `Self::active_ethereum_range().is_none()`?

pallets/eth-bridge/src/lib.rs
```diff
@@ -556,6 +549,69 @@ pub mod pallet {
...
+                });
+
+                let votes_count = sorted_votes
+                    .iter()
+                    .map(|(_range, votes)| votes)
+                    .fold(0, |acc, x| acc as usize + x);
+                let mut remaining_votes_threshold = AVN::<T>::supermajority_quorum() as usize;
+
+                if votes_count >= remaining_votes_threshold as usize {
+                    let quorum = AVN::<T>::quorum() as usize;
```
nahuseyoum, line 596: Why do we need 2 quorums?

pallets/eth-bridge/src/lib.rs
```diff
@@ -556,6 +549,69 @@ pub mod pallet {
...
+            if let Some(events_partition) =
+                events_helpers::discovered_eth_events_partition_factory(nominated_range, Vec::new())
+                    .first()
+            {
+                let mut votes = EthereumEvents::<T>::get(&events_partition);
+                votes.try_insert(author.account_id).map_err(|_| Error::<T>::EventVotesFull)?;
+
+                EthereumEvents::<T>::insert(&events_partition, votes);
+
+                let mut sorted_votes: Vec<(EthereumEventsPartition, usize)> =
```
nahuseyoum, line 581: If am not mistaken, we spoke about re-using existing storage items for the "initial" voting round but would it be simpler to reach consensus on the `initialEthereumBlock`, in its own storage (something that keeps track of who saw what block) until `quorum` number of collators have "seen" an ethereum block. Once they have, we can do the active range setup where `selected_range` starts from the `initialEthereumBlock`:
```
ActiveEthereumRange::<T>::put(ActiveEthRange {
                        range: selected_range,
                        partition: 0,
                        // TODO retrieve values from runtime.
                        event_types_filter: Default::default(),
                    });
```
We can even drain the special storage item we used to reach consensus on the initial voting round. What do you think?

pallets/eth-bridge/src/lib.rs
thadouk: it was fixed on the next PR but will backport here

pallets/eth-bridge/src/lib.rs
thadouk: I think its better if we use the existing data structures. The additional storage is practically dead code after the initial consensus that we need to keep and never use, regardless of it being drained or not.

pallets/eth-bridge/src/lib.rs
thadouk: we need a threshold of votes to start the consensus mechanism (supermajority_quorum) and then a `quorum` collators that have seen the same range

pallets/eth-bridge/src/lib.rs
thadouk: its a dummy implementation, see the TODO few lines bellow

pallets/eth-bridge/src/lib.rs
thadouk: its a factory method that produces partitions given an input of events. In the case of the initial range consensus we are interested in producing a partition with no events and the nominated_range

pallets/eth-bridge/src/lib.rs
thadouk: updated to use an independent storage.

PR #397, opened 2024-05-04
Integrate initial range voting with client and refactor runtime API

thadouk:
## Proposed changes
This commit includes the integration of the initial range voting functionality with the client.
It ports changes to integrate the runtime API for testing runtime behavior.
Additionally, it refactors the runtime API to improve proof creation by relocating helper
functions to primitive crates.
Fixes genesis config for test runtime.

Jira tasks:
- SYS-3930
- SYS-3635

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -7,7 +7,7 @@ use sc_keystore::LocalKeystore;
 use sp_api::ApiExt;
 use sp_avn_common::{
     event_discovery::{
-        create_proof_data, events_helpers::discovered_eth_events_partition_factory,
+        encode_eth_event_submission_data, events_helpers::discovered_eth_events_partition_factory,
         DiscoveredEvent, EthBlockRange, EthereumEventsPartition,
     },
     event_types::{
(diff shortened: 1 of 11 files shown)


ivan-cholakov (approved):

ivan-cholakov (approved):

primitives/avn-common/src/event_discovery.rs
```diff
@@ -120,6 +120,14 @@ impl EthereumEventsFilterTrait for () {
     }
 }
 
+pub fn create_proof_data<AccountId: Encode, Data: Encode>(
```
nahuseyoum, line 123:
```suggestion
pub fn encode_eth_event_submission_data<AccountId: Encode, Data: Encode>(
```

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -393,12 +394,107 @@ where
...
+        .map_err(|err| format!("Failed to check if author has casted latest  vote: {:?}", err))?;
+
+    if !has_casted_vote {
+        let web3_data_mutex = config.web3_data_mutex.lock().await;
+        let web3_ref = match web3_data_mutex.web3.as_ref() {
+            Some(web3) => web3,
+            None => return Err("Web3 connection not set up".into()),
+        };
+
+        let latest_seen_ethereum_block = web3_utils::get_current_block_number(web3_ref)
```
nahuseyoum, line 450: Should we get `current_block` - X to take into account finality?

nahuseyoum (approved):

MBrozhko34 (approved):

node/avn-service/src/ethereum_events_handler.rs
thadouk: I think there is no need because the value gets transformed into a range after reducing an amount of blocks, ensuring finality like that.
https://github.com/AventusProtocolFoundation/avn-parachain/blob/feat/SYS-3560-add-event-listener/primitives/avn-common/src/event_discovery.rs#L171-L179

PR #398, opened 2024-05-08
fix: refactored getting the authors to verify current node's author

ivan-cholakov:
instead of `getting the author for current node`, as it requires `externalities`, we are exposing the `validators` storage item to iterate through to find the correct keystore public key

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -360,6 +360,8 @@ where
     let author_public_key =
         config.client.runtime_api().query_authors(config.client.info().best_hash);
 
+    log::info!("HELP !!! AUTHOR: {:?}, PUBLIC: {:?}", author_public_key,public_keys);
+    
     let current_public_key = match find_author_account_id(author_public_key, public_keys) {
         Some(key) => key,
         None => {
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -19,7 +19,7 @@ use sp_avn_common::{
 };
 use sp_block_builder::BlockBuilder;
 use sp_blockchain::HeaderBackend;
-use sp_core::H256 as SpH256;
+use sp_core::{sr25519::Public, H256 as SpH256};
 use sp_keystore::Keystore;
 use sp_runtime::AccountId32;
 use std::{collections::HashMap, marker::PhantomData, time::Instant};
@@ -314,6 +314,21 @@ where
     Err(AppError::GenericError("Failed to initialize web3 after multiple attempts.".to_string()))
 }
 
+fn find_author_account_id<T>(
+    authors: Result<Vec<AccountId32>, T>,
+    keystore_public_keys: Vec<Public>,
+) -> Option<Public> {
+    if let Ok(account_ids) = authors {
+        for key in keystore_public_keys {
+            let keystore_account_id = AccountId32::from(key.0);
+            if account_ids.contains(&keystore_account_id) {
+                return Some(key)
+            }
+        }
+    }
+    None
+}
+
 pub async fn start_eth_event_handler<Block, ClientT>(config: EthEventHandlerConfig<Block, ClientT>)
 where
     Block: BlockT,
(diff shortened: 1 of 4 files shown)


node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -342,30 +357,20 @@ where
...
-        .find(|public_key| AccountId32::from(public_key.0) == author_public_key)
-    {
+    let author_public_key =
+        config.client.runtime_api().query_authors(config.client.info().best_hash)
+        .map_err(|e|{
+            log::error!("Error querying authors: {:?}", e);
+        })
+        .and_then(|opt_keys| match opt_keys {
+            Some(keys)=>Ok(keys),
+            None => todo!(),
```
nahuseyoum, line 367: Is this correct?

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -342,30 +357,20 @@ where
...
-            log::error!("{}", e);
-            return // Exit on error
-        },
-    };
-
-    let current_public_key = match public_keys
-        .into_iter()
-        .find(|public_key| AccountId32::from(public_key.0) == author_public_key)
-    {
+    let author_public_key =
```
nahuseyoum, line 360:
```suggestion
    let author_public_keys =
```

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -342,30 +357,20 @@ where
...
+        config.client.runtime_api().query_authors(config.client.info().best_hash)
+        .map_err(|e|{
+            log::error!("Error querying authors: {:?}", e);
+        })
+        .and_then(|opt_keys| match opt_keys {
+            Some(keys)=>Ok(keys),
+            None => todo!(),
+        });
+
+    let current_public_key = match find_author_account_id(author_public_key, public_keys) {
```
nahuseyoum, line 370:
```suggestion
    let current_node_public_key = match find_author_account_id(author_public_key, public_keys) {
```

runtime/avn/src/lib.rs
```diff
@@ -15,16 +15,21 @@ use core::cmp::Ordering;
 use codec::{Decode, Encode};
 use scale_info::TypeInfo;
 
+// use sp_application_crypto::RuntimePublic;
```
nahuseyoum, line 18:
```suggestion
```

runtime/avn/src/lib.rs
```diff
@@ -946,9 +957,14 @@ impl_runtime_apis! {
     }
 
     impl pallet_eth_bridge_runtime_api::EthEventHandlerApi<Block, AccountId> for Runtime {
-        fn query_current_author() -> Option<AccountId>{
-            Avn::get_validator_for_current_node()
-                .map(|validator| validator.account_id)
+        fn query_authors() ->Option<Vec<[u8; 32]>> {
```
nahuseyoum, line 960: I think we should rename this function to make it explicit that its returning the signing keys of the authors, not the authors themselves. The big difference here is that the authors (author address) cannot be changed but the signing key can be rotated. And unsigned transactions are signed using the signing key. So i would suggest renaming it to `query_author_signing_keys`

runtime/test/src/lib.rs
```diff
@@ -14,6 +14,10 @@ use core::cmp::Ordering;
 
 use codec::{Decode, Encode};
 use scale_info::TypeInfo;
+use sp_core::sr25519::Public;
+
+// use sp_application_crypto::RuntimePublic;
```
nahuseyoum, line 19:
```suggestion
```

runtime/test/src/lib.rs
```diff
@@ -966,9 +972,14 @@ impl_runtime_apis! {
     }
 
     impl pallet_eth_bridge_runtime_api::EthEventHandlerApi<Block, AccountId> for Runtime {
-        fn query_current_author() -> Option<AccountId>{
-            Avn::get_validator_for_current_node()
-                .map(|validator| validator.account_id)
+        fn query_authors() -> Option<Vec<[u8; 32]>>{
```
nahuseyoum, line 975: ditto

nahuseyoum (approved):

runtime/avn/src/lib.rs
```diff
@@ -177,7 +187,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("avn-parachain"),
     impl_name: create_runtime_str!("avn-parachain"),
     authoring_version: 1,
-    spec_version: 64,
+    spec_version: 60,
```
thadouk, line 190: why `60`?

PR #401, opened 2024-05-15
SYS-4051 Update weights for eth events

nahuseyoum:
## Proposed changes

This PR updates the weight calculation in Eth events pallet to use realistic values and decouple the max storage bound from the constants used in Weight calculation.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [x] Documentation updated


pallets/avn-offence-handler/src/lib.rs
```diff
@@ -50,7 +50,6 @@ pub mod pallet {
     }
 
     #[pallet::pallet]
-    #[pallet::generate_store(pub (super) trait Store)]
```
ivan-cholakov, line 53: This seems like a rebase is needed. I think Thanos removed this and it is already merged to main

ivan-cholakov (approved):

thadouk (approved):

PR #402, opened 2024-05-20
fix: client extrinsics operation

thadouk:
## Proposed changes
This commit addresses issues with extrinsic submission in the runtime:

- Registers the extension to the instance submitting the extrinsic.
- Updates signature validation.
- Simplifies interface and error handling via the runtime API, following the report_equivocation style.

### DIFF
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -393,6 +393,7 @@ impl<T: Config> Pallet<T> {
     ) -> bool {
         // verify that the incoming (unverified) pubkey is actually a validator
         if !Self::is_validator(&validator.account_id) {
+            warn!("Signature validation failed, account {:?}, is not validator", validator);
             return false
         }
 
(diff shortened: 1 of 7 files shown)


pallets/avn/src/lib.rs
```diff
@@ -393,13 +393,21 @@ impl<T: Config> Pallet<T> {
...
+            warn!("Signature validation failed, account {:?}, is not validator", validator);
             return false
         }
 
         // check signature (this is expensive so we do it last).
         let signature_valid =
             data.using_encoded(|encoded_data| validator.key.verify(&encoded_data, &signature));
 
+        debug!(
+            "🪲 Validating signature: [ data {:?} - account {:?} - signature {:?} ] Result: {}",
```
nahuseyoum, line 405: is the emoji a `bug`? :smile:

pallets/eth-bridge/src/lib.rs
```diff
@@ -894,7 +894,7 @@ pub mod pallet {
                     },
                 Call::submit_ethereum_events { author, events_partition, signature } =>
                     if AVN::<T>::signature_is_valid(
-                        &encode_eth_event_submission_data(
+                        &(
```
nahuseyoum, line 897: why are we removing the function call? Won't this potentially cause a mistmatch?

nahuseyoum (approved):

pallets/avn/src/lib.rs
thadouk: cause it is **de**-bug level of info :tada:

pallets/eth-bridge/src/lib.rs
thadouk: its a missmatch now. Validation expects a tupple, not encoded data.

PR #403, opened 2024-05-21
SYS-4051 Update weight constants

nahuseyoum:
## Proposed changes

This PR changes the EthereumEvents constants to allow creating multiple lifts (more than 4) to be sent in a batch transaction. This is needed by the lift sweeper to send lift transactions to the chain.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [x] Documentation updated


thadouk (approved):
The weight changes seem small. I was expecting significant reduce on the cost of the add ethereum events extrinsics

nahuseyoum:
> The weight changes seem small. I was expecting significant reduce on the cost of the add ethereum events extrinsics

the base weight went down from `761,487` to  `63,886` which seems to be pretty significant. We can also now send 57 lifts in a batch instead of 5.

micaelffrancoAV (approved):

PR #404, opened 2024-05-23
fix: Make avn-event-handler service more faullt tollerant

thadouk:
This commit allows the avn-event-hanlder to handle more graceful error.
Also applies code formatting.

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -349,27 +349,29 @@ where
 
     log::info!("⛓️  ETH EVENT HANDLER INITIALIZED");
 
-    let public_keys = config.keystore.sr25519_public_keys(AVN_KEY_ID);
+    let mut current_node_public_key = Public([0u8; 32]);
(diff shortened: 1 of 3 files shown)


node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -349,27 +349,29 @@ where
...
-
-    let current_node_public_key = match find_author_account_id(author_public_keys, public_keys) {
-        Some(key) => key,
-        None => {
-            log::error!("Author not found");
-            return
-        },
-    };
+    let mut current_node_public_key = Public([0u8; 32]);
+    loop {
```
nahuseyoum, line 353: Am not sure if we want to loop indefinitly. Perhaps a set number of retries should be done before we give us and resume operations without throwing an error and killing the node?

thadouk: If its not busy waiting, and this is an essential task for the operation of a collator, why shouldn't we loop till we achieve the task?

PR #408, opened 2024-05-30
SYS-4089 Benchmark for new extrinsics

nahuseyoum:


fluorostani (approved):

thadouk (approved):
Why do we need two bench per extrinsic?

ivan-cholakov (approved):

PR #410, opened 2024-05-31
SYS-4072 fix handling of eth-bridge processed events

thadouk:
## Proposed changes

Adds handlers in Token manager and wire up for Runtime to process the first type of events from eth-bridge.

Jira ticket:
- SYS-4072

### DIFF
--- a/pallets/token-manager/src/lib.rs
+++ b/pallets/token-manager/src/lib.rs
@@ -1092,4 +1092,15 @@ impl<T: Config> BridgeInterfaceNotification for Pallet<T> {
 
         Ok(())
     }
+
+    fn on_incoming_event_processed(event: &EthEvent) -> DispatchResult {
+        match &event.event_data {
+            EventData::LogLifted(d) => return Self::process_lift(event, d),
+            EventData::LogAvtGrowthLifted(d) => return Self::process_avt_growth_lift(event, d),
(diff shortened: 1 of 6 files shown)


pallets/token-manager/src/lib.rs
```diff
@@ -1092,4 +1092,15 @@ impl<T: Config> BridgeInterfaceNotification for Pallet<T> {
 
         Ok(())
     }
+
+    fn on_incoming_event_processed(event: &EthEvent) -> DispatchResult {
+        match &event.event_data {
+            EventData::LogLifted(d) => return Self::process_lift(event, d),
+            EventData::LogAvtGrowthLifted(d) => return Self::process_avt_growth_lift(event, d),
```
nahuseyoum, line 1099: Would it be "safer" to add these once we add the filters to the runtime and remove them from eth-events?

nahuseyoum (approved):

ivan-cholakov (approved):

pallets/token-manager/src/lib.rs
thadouk: can do, but the filter is expected to define and constrain the event types the runtime process. The pallets should just be processing events as they come through the interface.

fluorostani (approved):

PR #415, opened 2024-06-06
SYS-3560 add error on rejected event

nahuseyoum:


nahuseyoum:
Wrong parent branch used

PR #419, opened 2024-06-12
SYS-3635 deprecate ethreum events entrypoints

thadouk:
## Proposed changes

Deprecates ethereum-events entry points that add manually events in the system.
Moving forward the responsibility will be shifted to eth-bridge event scanning mechanism on Ethereum.

### DIFF
--- a/pallets/ethereum-events/src/lib.rs
+++ b/pallets/ethereum-events/src/lib.rs
@@ -425,7 +425,8 @@ pub mod pallet {
         // We need to maintain this till SYS-888 is resolved. After that it can be removed.
         #[deprecated(
             since = "5.5.0",
-            note = "This extrinsic is being deprecated. Use add_ethereum_log"
+            note = "This extrinsic is being deprecated and will be removed in the near future. Ethereum events will be automatically imported by EthBridge pallet.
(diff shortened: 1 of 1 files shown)


ivan-cholakov (approved):

pallets/ethereum-events/src/lib.rs
```diff
@@ -422,8 +422,11 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
-        /// This extrinsic is being deprecated. Use add_ethereum_log
         // We need to maintain this till SYS-888 is resolved. After that it can be removed.
+        #[deprecated(
+            since = "5.5.0",
+            note = "This extrinsic is being deprecated. Use add_ethereum_log"
```
nahuseyoum, line 428: Will this cause a circular reference :smile:? Since add_ethereum_log is also being deprecated? Maybe we should have the same deprecation note as add_ethereum_log?

nahuseyoum (approved):

pallets/ethereum-events/src/lib.rs
thadouk: I can update the message, to reflect that they are both deprecated

PR #421, opened 2024-06-18
fix: event parsing and adds test case

thadouk:
## Proposed changes

Adds AvtLowerClaimedData in the types the client scans. Fixes the way the data field is constructed ensuring the data field is explicitly set as None when no data is present, to align with the runtime implementation.

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -11,9 +11,9 @@ use sp_avn_common::{
         DiscoveredEvent, EthBlockRange, EthereumEventsPartition,
     },
     event_types::{
-        AddedValidatorData, AvtGrowthLiftedData, Error, EthEvent, EthEventId, EventData,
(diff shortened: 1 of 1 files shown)


node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -197,8 +207,14 @@ fn parse_log(log: Log, events_registry: &EventRegistry) -> Result<DiscoveredEven
     let event_id = EthEventId { signature, transaction_hash: SpH256::from(transaction_hash.0) };
 
     let topics: Vec<Vec<u8>> = log.topics.iter().map(|t| t.0.to_vec()).collect();
-    let event_data = parse_event_data(signature, log.data.0, topics, events_registry)
-        .or_else(|_| Err(AppError::ErrorParsingEventLogs))?;
+    let data: Option<Vec<u8>> = if log.data.0.is_empty() { None } else { Some(log.data.0) };
```
ivan-cholakov, line 210: :100:

ivan-cholakov (approved):

PR #430, opened 2024-08-21
Sys 4191 register chain handler

ivan-cholakov:
Initial implementation of `register_chain_handler` and `update_chain_handler`

### DIFF
--- a/pallets/avn-anchor/src/lib.rs
+++ b/pallets/avn-anchor/src/lib.rs
@@ -5,7 +5,6 @@ extern crate alloc;
 #[cfg(not(feature = "std"))]
 use alloc::string::String;
 
-// use pallet_avn::{self as avn};
 use frame_support::{
     dispatch::DispatchResult, ensure, pallet_prelude::StorageVersion, traits::Get,
 };
--- a/pallets/avn-anchor/src/lib.rs
+++ b/pallets/avn-anchor/src/lib.rs
@@ -6,18 +6,19 @@ extern crate alloc;
 use alloc::string::String;
 
 // use pallet_avn::{self as avn};
-use frame_system::{
-    self as system, ensure_none, ensure_root,
-};
 use frame_support::{
     dispatch::DispatchResult, ensure, pallet_prelude::StorageVersion, traits::Get,
 };
+use frame_system::{self as system, ensure_none, ensure_root};
 // use sp_std::prelude::*;
 
 pub mod default_weights;
 pub use default_weights::WeightInfo;
 
 pub use pallet::*;
+use sp_core::ConstU32;
+
+pub type MaximumHandlersBound = ConstU32<256>;
 
 #[frame_support::pallet]
 pub mod pallet {
@@ -25,12 +26,13 @@ pub mod pallet {
     use frame_support::pallet_prelude::*;
     use frame_system::pallet_prelude::*;
 
+    pub type ChainId = u32;
+
     #[pallet::config]
-    pub trait Config: frame_system::Config + pallet_avn::Config 
-    {
-        // type RuntimeEvent: From<Event<Self>>
-        //     + Into<<Self as frame_system::Config>::RuntimeEvent>
-        //     + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+    pub trait Config: frame_system::Config + pallet_avn::Config {
+        type RuntimeEvent: From<Event<Self>>
+            + Into<<Self as frame_system::Config>::RuntimeEvent>
+            + IsType<<Self as frame_system::Config>::RuntimeEvent>;
 
         type WeightInfo: WeightInfo;
     }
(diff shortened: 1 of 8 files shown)


pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,110 @@ extern crate alloc;
...
+    pub type ChainHandlers<T: Config> = StorageMap<_, Blake2_128Concat, ChainId, T::AccountId>;
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
         #[pallet::weight(<T as pallet::Config>::WeightInfo::register_chain_handler())]
         #[pallet::call_index(0)]
         pub fn register_chain_handler(
             origin: OriginFor<T>,
-            // handler:
+            chain_id: ChainId,
```
thadouk, line 76: I would suggest the input here is a friendly name, and ChainId is a nonce managed by the pallet, ensuring uniqueness 
We store the chain friendly name with the pallet data.

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,110 @@ extern crate alloc;
...
         #[pallet::call_index(1)]
         pub fn update_chain_handler(
             origin: OriginFor<T>,
-            // handler:
+            chain_id: ChainId,
+            new_handler_account_id: T::AccountId,
         ) -> DispatchResult {
+            ensure_signed(origin)?;
+
+            ChainHandlers::<T>::try_mutate(&chain_id, |maybe_handler| -> DispatchResult {
```
thadouk, line 102: We need to add a check here to ensure that the sender of this is actually the handler account otherwise return an error.

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,110 @@ extern crate alloc;
...
+                let handler = maybe_handler.as_mut().ok_or(Error::<T>::HandlerNotRegistered)?;
```
thadouk, line 103: I think if the data structure is chain_id -> handler if the chain id is not returning a handler then we should error with something that says that "chain_id is not registered" rather that the handler was not found.

pallets/avn-anchor/src/tests.rs
```diff
@@ -0,0 +1,144 @@
...
+            AvnAnchor::update_chain_handler(
+                RuntimeOrigin::signed(new_handler),
+                chain_id,
+                new_handler
+            ),
+            Error::<TestRuntime>::HandlerNotRegistered
+        );
+    });
+}
+
```
thadouk, line 118: we also need a test here that tests that a handler that attempts an update is possible only if the sender is the existing handler

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,129 @@ extern crate alloc;
 use alloc::string::String;
 
 // use pallet_avn::{self as avn};
```
nahuseyoum, line 8:
```suggestion
```

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,129 @@ extern crate alloc;
...
         #[pallet::weight(<T as pallet::Config>::WeightInfo::register_chain_handler())]
         #[pallet::call_index(0)]
         pub fn register_chain_handler(
             origin: OriginFor<T>,
-            // handler:
+            name: BoundedVec<u8, ChainNameLimit>,
         ) -> DispatchResult {
+            let handler = ensure_signed(origin)?;
+
+            ensure!(!ChainHandlers::<T>::contains_key(&handler), Error::<T>::HandlerAlreadyRegistered);
```
nahuseyoum, line 94: Can we validate that `name` is not null?

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,129 @@ extern crate alloc;
...
-            // handler:
+            new_handler: T::AccountId,
         ) -> DispatchResult {
+            let old_handler = ensure_signed(origin)?;
+
+            let chain_data = ChainHandlers::<T>::get(&old_handler).ok_or(Error::<T>::ChainNotRegistered)?;
+
+            ensure!(!ChainHandlers::<T>::contains_key(&new_handler), Error::<T>::HandlerAlreadyRegistered);
+
+            ChainHandlers::<T>::remove(&old_handler);
```
nahuseyoum, line 122: Can't we use `mutate`?

pallets/avn-anchor/src/tests.rs
```diff
@@ -0,0 +1,143 @@
...
+        assert_eq!(chain_data1.chain_id, 0);
+        assert_eq!(chain_data1.name, name1);
+        assert_eq!(chain_data2.chain_id, 1);
+        assert_eq!(chain_data2.name, name2);
+
+        System::assert_has_event(Event::ChainHandlerRegistered(handler1, 0, name1).into());
+        System::assert_has_event(Event::ChainHandlerRegistered(handler2, 1, name2).into());
+    });
+}
```
nahuseyoum, line 143: Can we add a test for null chain name?

nahuseyoum (approved):

pallets/avn-anchor/src/lib.rs
```diff
@@ -6,62 +6,129 @@ extern crate alloc;
...
             Ok(())
         }
 
-        #[pallet::weight( <T as pallet::Config>::WeightInfo::update_chain_handler())]
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::update_chain_handler())]
         #[pallet::call_index(1)]
         pub fn update_chain_handler(
             origin: OriginFor<T>,
-            // handler:
+            new_handler: T::AccountId,
```
thadouk, line 114: Do you think that we should have a proof attached here with signed data of the new_handler?

thadouk (approved):

pallets/avn-anchor/src/lib.rs
ivan-cholakov: Not sure this is needed since the only actor that can set the new handler is the current handler

PR #431, opened 2024-08-21
Sys 4193 submit checksum with identity

ivan-cholakov:
Added `submit_checkpoint_with_identity` extrinsic


pallets/avn-anchor/src/lib.rs
```diff
@@ -55,18 +49,38 @@ pub mod pallet {
...
+        Blake2_128Concat,
+        CheckpointId,
+        H256,
+        ValueQuery,
+    >;
+
+    #[pallet::storage]
+    #[pallet::getter(fn next_checkpoint_id)]
+    pub type NextCheckpointId<T: Config> =
+        StorageMap<_, Blake2_128Concat, ChainId, CheckpointId, ValueQuery>;
```
thadouk, line 82: why do we need a unique checkpoint_id per chain? Can it not be a universal nonce for simplicity?

pallets/avn-anchor/src/lib.rs
```diff
@@ -81,6 +80,22 @@ pub mod pallet {
...
+        ChainId,
+        Blake2_128Concat,
+        CheckpointId,
+        H256,
+        ValueQuery,
+    >;
+
+    #[pallet::storage]
+    #[pallet::getter(fn next_checkpoint_id)]
+    pub type NextCheckpointId<T> = StorageValue<_, CheckpointId, ValueQuery>;
```
nahuseyoum, line 97: Do we want this to be global to the pallet? Doesn't it make more sense it it was scoped to the ChainId?

pallets/avn-anchor/src/lib.rs
```diff
@@ -132,13 +151,28 @@ pub mod pallet {
             Ok(())
         }
 
-        #[pallet::weight( <T as pallet::Config>::WeightInfo::submit_checkpoint_with_identity())]
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::submit_checkpoint_with_identity())]
```
nahuseyoum, line 154: In a separate PR, can we create a `signed` version of this function please? We will need it to allow users to call this via the Gateway

PR #433, opened 2024-09-11
SYS-4294 Wireup with avnProxy

nahuseyoum:


pallets/avn-anchor/src/lib.rs
```diff
@@ -314,25 +311,6 @@ pub mod pallet {
 
             Ok(())
         }
-
-        #[pallet::call_index(6)]
-        #[pallet::weight(<T as pallet::Config>::WeightInfo::proxy().saturating_add(call.get_dispatch_info().weight))]
-        pub fn proxy(
```
ivan-cholakov, line 320: Why are you removing the proxy extrinsic?

PR #435, opened 2024-09-16
SYS-4189 multi currency support

nahuseyoum:
## Proposed changes

This PR enables support for multi currency payment when sending transactions via the Gateway. This means users will be able to pay Relayers in a variety of ERC20 tokens.

Tests and Benchmarks are also updated to be compatible.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [x] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [x] Documentation updated


pallets/avn-proxy/src/lib.rs
```diff
@@ -211,7 +233,7 @@ impl<T: Config> Pallet<T> {
 
     pub(crate) fn charge_fee(
         proof: &Proof<T::Signature, T::AccountId>,
-        payment_info: PaymentInfo<T::AccountId, BalanceOf<T>, T::Signature>,
+        payment_info: PaymentInfo<T::AccountId, BalanceOf<T>, T::Signature, T::Token>,
```
thadouk, line 236: can this be abused? meaning any ERC20 token can be used as a payment proof?

thadouk (approved):
Overall nice work. Is it possible to avoid the hardcoding of mnemonics and the dependency between accounts and genesis.

pallets/avn-proxy/src/lib.rs
nahuseyoum: This is a good question. If someone sends a transaction direct to chain, and decides to pay for it, it will be done voluntarily (the whole payment_info is optional). The real check to make sure `supported` tokens are used as payment is done in the gateway by the relayer.

PR #438, opened 2024-09-22
SYS-4220 Disable lowering

nahuseyoum:
## Proposed changes

This PR introduces a flag to toggle lowering capabilities of the chain.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


pallets/token-manager/src/lib.rs
```diff
@@ -313,6 +316,11 @@ pub mod pallet {
     #[pallet::getter(fn lower_schedule_period)]
     pub type LowerSchedulePeriod<T: Config> = StorageValue<_, BlockNumberFor<T>, ValueQuery>;
 
+    /// The number of blocks lower transactions are delayed before executing
```
micaelffrancoAV, line 319: Is this comment for this storage item?

micaelffrancoAV (approved):

fluorostani (approved):

pallets/token-manager/src/lib.rs
```diff
@@ -574,6 +586,22 @@ pub mod pallet {
 
             return Ok(())
         }
+
+        #[pallet::call_index(10)]
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::toggle_lowering())]
+        pub fn toggle_lowering(origin: OriginFor<T>, enabled: bool) -> DispatchResult {
```
thadouk, line 592: I prefer having a `set_config` function where different values can be configured
we already have `set_lower_schedule_period` so it would be better if we have one set admin setting function.

pallets/token-manager/src/lib.rs
```diff
@@ -238,6 +238,8 @@ pub mod pallet {
         LowerSchedulePeriodUpdated {
             new_period: BlockNumberFor<T>,
         },
+        LoweringEnabled,
+        LoweringDisabled,
```
thadouk, line 242: One event for settings is better to avoid bloating the pallet codebase.

pallets/token-manager/src/lib.rs
nahuseyoum: I personally prefer having smaller extrinsics that do specific tasks. They are easier to test, benchmark and reason with. Happy to hear other thoughts

pallets/token-manager/src/lib.rs
nahuseyoum: I personally prefer having specific events that describe exactly what is happening instead of a generic one with different argument. Again happy to hear what others think.

pallets/token-manager/src/lib.rs
micaelffrancoAV: For me, with the little experience I have, I feel like smaller extrinsics are easier to understand but happy either way!

thadouk: I'm afraid I have to disagree, multiple entry points for sudo configuration is a security concern.
The codebase increases significantly if you have to replicate all origin and benchmark tests for each one of the entrypoints, making it difficult to navigate through files.
This is a common practice in many pallet.
Alternatively we can make this a const, defined by the runtime. Its clean and minimal and the way many pallets are configured in substrate.

thadouk (approved):
The functionality of the code introduced on this PR is correct, the approach to changing the configuration should improve as mentioned.
Although this has an overhead, this can be done on a separate refactor since the functionality needs to be expedited.
Approved on that basis.

PR #440, opened 2024-09-27
feat: Enhance avn-service configuration and logging

thadouk:
## Proposed changes

Improves avn-service checks to ensure signing operations are only performed when the node is correctly configured with a single Ethereum key in the keystore.

Additionally, enhances logging around signing operations for better visibility and eliminates multiple implementations of the ethereum hashing function.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/node/avn-service/src/keystore_utils.rs
+++ b/node/avn-service/src/keystore_utils.rs
@@ -20,6 +20,13 @@ pub fn get_eth_address_bytes_from_keystore(keystore_path: &PathBuf) -> Result<Ve
         Err(server_error(format!("No keys found in the keystore for {:?}", ETHEREUM_SIGNING_KEY)))?
     }
 
+    if addresses.len() > 1 {
+        Err(server_error(format!(
+            "Multiple keys found in the keystore for {:?}. Only one should be present.",
+            ETHEREUM_SIGNING_KEY
+        )))?
+    }
+
     return Ok(addresses[0].clone())
 }
 
--- a/node/avn-service/src/lib.rs
+++ b/node/avn-service/src/lib.rs
@@ -5,10 +5,10 @@ use jsonrpc_core::ErrorCode;
 use sc_client_api::{client::BlockBackend, UsageProvider};
 use sc_keystore::LocalKeystore;
 use sp_avn_common::{
-    EthQueryRequest, EthQueryResponse, EthQueryResponseType, EthTransaction,
-    DEFAULT_EXTERNAL_SERVICE_PORT_NUMBER,
+    hash_with_ethereum_prefix, EthQueryRequest, EthQueryResponse, EthQueryResponseType,
+    EthTransaction, DEFAULT_EXTERNAL_SERVICE_PORT_NUMBER,
 };
-use sp_core::{ecdsa::Signature, hashing::keccak_256};
+use sp_core::ecdsa::Signature;
 use sp_runtime::traits::Block as BlockT;
 use std::{marker::PhantomData, time::Instant};
 use web3::{transports::Http, types::TransactionReceipt, Web3};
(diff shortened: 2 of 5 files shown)


node/avn-service/src/keystore_utils.rs
```diff
@@ -20,6 +20,13 @@ pub fn get_eth_address_bytes_from_keystore(keystore_path: &PathBuf) -> Result<Ve
         Err(server_error(format!("No keys found in the keystore for {:?}", ETHEREUM_SIGNING_KEY)))?
     }
 
+    if addresses.len() > 1 {
```
nahuseyoum, line 23: :+1:

pallets/avn/src/lib.rs
```diff
@@ -417,18 +417,32 @@ impl<T: Config> Pallet<T> {
     ) -> bool {
         // verify that the incoming (unverified) pubkey is actually a validator
         if !Self::is_validator(&validator.account_id) {
+            log::warn!("✋ Account: {:?} is not a authority.", &validator.account_id);
```
nahuseyoum, line 420:
```suggestion
            log::warn!("✋ Account: {:?} is not an authority.", &validator.account_id);
```

primitives/avn-common/src/lib.rs
```diff
@@ -173,13 +173,20 @@ pub fn recover_public_key_from_ecdsa_signature(
...
+pub fn hash_with_ethereum_prefix(hex_message: &String) -> Result<[u8; 32], ECDSAVerificationError> {
     let message_bytes = hex::decode(hex_message.trim_start_matches("0x"))
         .map_err(|_| ECDSAVerificationError::InvalidMessageFormat)?;
 
     let mut prefixed_message = ETHEREUM_PREFIX.to_vec();
     prefixed_message.append(&mut message_bytes.to_vec());
-    Ok(keccak_256(&prefixed_message))
+    let hash = keccak_256(&prefixed_message);
+    log::debug!(
+        "🪲 Request to sign data: {:?},\n data with ethereum prefix: {:?}, \n result hash: {:?}",
```
nahuseyoum, line 184:
```suggestion
        "🪲 Data without prefix: {:?},\n data with ethereum prefix: {:?}, \n result hash: {:?}",
```

primitives/avn-common/src/lib.rs
```diff
@@ -173,13 +173,20 @@ pub fn recover_public_key_from_ecdsa_signature(
     }
 }
 
-pub fn hash_with_ethereum_prefix(hex_message: String) -> Result<[u8; 32], ECDSAVerificationError> {
+pub fn hash_with_ethereum_prefix(hex_message: &String) -> Result<[u8; 32], ECDSAVerificationError> {
```
nahuseyoum, line 176: If we are centralising I think we should use the version that accepts bytes (the one in avn-service) and replace this function with that. The conversion `toString()` might affect the bytes so its safer to work with bytes.

node/avn-service/src/lib.rs
```diff
@@ -428,16 +421,15 @@ where
             let secp = Secp256k1::new();
             let keystore_path = &req.state().keystore_path;
 
-            let data_to_sign: Vec<u8> =
-                hex::decode(req.param("data_to_sign")?.trim_start_matches("0x")).map_err(|e| {
+            let data_to_sign = req.param("data_to_sign")?;
+            let hashed_message =
+                hash_with_ethereum_prefix(&data_to_sign.to_string()).map_err(|e| {
```
nahuseyoum, line 426: what happened with the `trim`?

nahuseyoum (approved):
Nice!

primitives/avn-common/src/lib.rs
thadouk: well both convert string to bytes, trim `0x` prefix.
The one in avn-service does it outside the function, the one in avn-common does it bundled in a fucntion.
So it makes more sense to keep this version.
That is till we replace it all together with the one provided by ethers-rs implementation

node/avn-service/src/lib.rs
thadouk: its happening in the `hash_with_ethereum_prefix` function

ivan-cholakov (approved):

PR #443, opened 2024-10-11
SYS-4254 Summary pallet change feature merge

nahuseyoum:
## Proposed changes

This PR contains all the summary pallet changes (2 separate PRs)

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/pallets/summary/src/lib.rs
+++ b/pallets/summary/src/lib.rs
@@ -1535,4 +1535,8 @@ mod tests_set_periods;
 #[path = "tests/test_ocw_locks.rs"]
 mod test_ocw_locks;
 
+#[cfg(test)]
+#[path = "tests/anchor_tests.rs"]
+mod anchor_tests;
+
 // TODO: Add unit tests for setting schedule period and voting period
--- a/pallets/summary/src/lib.rs
+++ b/pallets/summary/src/lib.rs
@@ -52,7 +52,7 @@ use crate::offence::{create_and_report_summary_offence, SummaryOffence, SummaryO
 
 pub type EthereumTransactionId = u32;
 
-const PALLET_ID: &'static [u8; 7] = b"summary";
+const PALLET_ID: &'static [u8; 8] = b"summary-";
 const UPDATE_BLOCK_NUMBER_CONTEXT: &'static [u8] = b"update_last_processed_block_number";
 const ADVANCE_SLOT_CONTEXT: &'static [u8] = b"advance_slot";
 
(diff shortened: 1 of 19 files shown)


ivan-cholakov (approved):

pallets/summary/src/lib.rs
```diff
@@ -127,8 +127,12 @@ pub mod pallet {
 
         /// Weight information for the extrinsics in this pallet.
         type WeightInfo: WeightInfo;
-
+        /// An Ethereum bridge provider
         type BridgeInterface: avn::BridgeInterface;
+        /// A flag to determine if summaries will be automatically sent to Ethereum
+        type AutoSubmitSummaries: Get<bool>;
+        /// A unique instance id to differentiate different instances
+        type InstanceId: Get<u8>;
```
thadouk, line 135: The `AutoSubmitSummaries` and `InstanceId` could be integrated into the `BridgeInterface` Trait, making it more explicit.

pallets/summary/src/lib.rs
```diff
@@ -318,6 +322,16 @@ pub mod pallet {
...
 
+    #[pallet::storage]
+    #[pallet::getter(fn anchor_roots_counter)]
+    pub type AnchorRootsCounter<T: Config<I>, I: 'static = ()> = StorageValue<_, u32, ValueQuery>;
+
+    // Roots created to be anchored to other chains (apart from Ethereum)
+    #[pallet::storage]
+    #[pallet::getter(fn anchor_roots)]
+    pub type AnchorRoots<T: Config<I>, I: 'static = ()> =
+        StorageMap<_, Blake2_128Concat, u32, H256, ValueQuery>;
```
thadouk, line 333: why do we need a separate set of storage items? the pallets original storage items should be enough.

thadouk (approved):

pallets/summary/src/lib.rs
nahuseyoum: `BridgeInterface` shouldn't need to know the business logic of its calling pallets. It responsibility is to either send a transaction to Ethereum or generate a lower proof.

pallets/summary/src/lib.rs
nahuseyoum: The previous data structure was not designed to allow easy access for roots after they have been generated. Since the anchoring service needs to do this frequently and db read/write is not cheap and looping through all entries of a map is not sustainable, this new storage item solves these problems.

PR #448, opened 2024-10-24
Sys 4256 anchoring fees

ivan-cholakov:
## Proposed changes

Added fees to `avn-anchor` pallet, updated unit tests, updated benchmarks

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [x] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


pallets/avn-anchor/src/lib.rs
```diff
@@ -111,6 +148,17 @@ pub mod pallet {
...
 
+    #[pallet::storage]
+    #[pallet::getter(fn checkpoint_fee)]
+    pub type CheckpointFee<T> = StorageMap<_, Blake2_128Concat, ChainId, BalanceOf<T>, ValueQuery>;
+
+    /// An account nonce that represents the number of payments from this account
+    /// It is shared for all proxy transactions performed by that account
+    #[pallet::storage]
+    #[pallet::getter(fn payment_nonces)]
+    pub type PaymentNonces<T: Config> =
```
nahuseyoum, line 159: Do we need this? The handler nonce should be enough I think.

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +336,57 @@ pub mod pallet {
...
+            CheckpointFee::<T>::insert(chain_id, new_fee);
+            Self::deposit_event(Event::CheckpointFeeUpdated { chain_id, new_fee });
+
+            Ok(())
+        }
     }
 
     impl<T: Config> Pallet<T> {
+        pub(crate) fn charge_fee(handler: T::AccountId, chain_id: ChainId) -> DispatchResult {
+            let payment_nonce = PaymentNonces::<T>::get(&handler);
```
nahuseyoum, line 358: ditto

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +336,57 @@ pub mod pallet {
...
+            let checkpoint_fee = Self::get_checkpoint_fee(chain_id);
+            let treasury_account_id = compute_treasury_account_id::<T>();
+
+            T::FeeHandler::pay_fee(
```
nahuseyoum, line 362: I think we should have a function called `pay_treasury` in `FeeHander` and avoid passing in a `TreasuryAccount` and deriving an accountID from it. In case we ever change how we derive that account in TokenManager this would diverge without us knowing about it.

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +336,57 @@ pub mod pallet {
...
+                chain_id,
+            });
+
+            Ok(())
+        }
+
+        pub fn get_checkpoint_fee(chain_id: ChainId) -> BalanceOf<T> {
+            let fee = CheckpointFee::<T>::get(chain_id);
+            if fee == BalanceOf::<T>::default() {
+                T::DefaultCheckpointFee::get()
```
nahuseyoum, line 384: Would it be more explicit if it was an option query or if we were checking for BalanceOf::<T>::zero()?

pallets/avn-anchor/src/lib.rs
```diff
@@ -78,6 +84,27 @@ pub mod pallet {
...
+        /// A handler to process relayer fee payments
+        type FeeHandler: FeePaymentHandler<
+            AccountId = Self::AccountId,
+            Token = Self::Token,
+            TokenBalance = <Self::Currency as Currency<Self::AccountId>>::Balance,
+            Error = DispatchError,
+        >;
+
+        /// The account that will receive the checkpoint fees
+        type TreasuryAccount: Get<PalletId>;
```
nahuseyoum, line 104: I think we can remove it. See comment below

runtime/avn/src/lib.rs
```diff
@@ -679,6 +684,7 @@ parameter_types! {
     pub const StringLimit: u32 = 50;
     pub const MetadataDepositBase: Balance = 1 * MILLI_AVT;
     pub const MetadataDepositPerByte: Balance = 100 * MICRO_AVT;
+    pub const DefaultCheckpointFee: Balance = 1 * AVT;
```
nahuseyoum, line 687:
```suggestion
    pub const DefaultCheckpointFee: Balance = 0.1 * AVT;
```

pallets/avn-anchor/src/lib.rs
ivan-cholakov: we want to be able to set the fee to 0 right?

pallets/avn-anchor/src/lib.rs
```diff
@@ -515,3 +584,5 @@ pub fn encode_signed_submit_checkpoint_params<T: Config>(
 pub fn get_chain_data_for_handler<T: Config>(handler: &T::AccountId) -> Option<ChainDataStruct> {
     Pallet::<T>::chain_handlers(handler).and_then(|chain_id| Pallet::<T>::chain_data(chain_id))
 }
+
+use sp_core::Get;
```
nahuseyoum, line 588: Can we move this with the rest of the imports?

pallets/avn-anchor/src/mock.rs
```diff
@@ -31,9 +89,29 @@ frame_support::construct_runtime!(
         Avn: pallet_avn::{Pallet, Storage, Event},
         AvnProxy: avn_proxy::{Pallet, Call, Storage, Event<T>},
         AvnAnchor: avn_anchor::{Pallet, Call, Storage, Event<T>},
+        TokenManager: pallet_token_manager::{Pallet, Call, Storage, Event<T>},
```
nahuseyoum, line 92: it must have been fun to implement all these :smile:

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +319,47 @@ pub mod pallet {
...
+
+            Ok(())
+        }
     }
 
     impl<T: Config> Pallet<T> {
+        pub(crate) fn charge_fee(handler: T::AccountId, chain_id: ChainId) -> DispatchResult {
+            let checkpoint_fee = Self::get_checkpoint_fee(chain_id);
+
+            T::FeeHandler::pay_treasury(&T::Token::default(), &checkpoint_fee, &handler)?;
```
nahuseyoum, line 343: Do we want to pass in a default here?

pallets/avn-anchor/src/lib.rs
nahuseyoum: Yes, IMO the correct way to handle that would be to make the storage an OptionsQuery instead of a ValueQuery item. Unless I am mistaken, `BalanceOf::<T>::default()` returns 0 which would make this use the default fee.

pallets/parachain-staking/src/tests/mock.rs
```diff
@@ -693,6 +693,13 @@ impl FeePaymentHandler for Test {
     ) -> Result<(), Self::Error> {
         return Err(DispatchError::Other("Test - Error"))
     }
+    fn pay_treasury(
```
nahuseyoum, line 696: Do we need this? I thought we had an implementation for `()`

pallets/ethereum-events/src/mock.rs
```diff
@@ -835,4 +835,12 @@ impl FeePaymentHandler for TestRuntime {
     ) -> Result<(), Self::Error> {
         return Err(DispatchError::Other("Test - Error"))
     }
+
+    fn pay_treasury(
```
nahuseyoum, line 839: Do we need this? I thought we had an implementation for `()`

pallets/avn-anchor/src/tests.rs
```diff
@@ -614,13 +600,140 @@ fn checkpoint_id_overflow_fails() {
...
+        assert_ok!(AvnAnchor::charge_fee(handler.clone(), chain_id));
+
+        System::assert_last_event(
+            Event::CheckpointFeeCharged { handler: handler.clone(), chain_id, fee }.into(),
+        );
+    });
+}
+
+#[test]
+fn submit_checkpoint_charges_correct_fee() {
```
nahuseyoum, line 658: Can we add a test to check if setting a 0 fee works?

pallets/avn-anchor/src/mock.rs
ivan-cholakov: :D

pallets/avn-anchor/src/lib.rs
```diff
@@ -139,6 +173,10 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    #[pallet::storage]
+    #[pallet::getter(fn fee_token)]
+    pub type FeeToken<T: Config> = StorageValue<_, T::Token, ValueQuery>;
```
nahuseyoum, line 178: If we do it this way, we have to have a genesis config and a state migration to set it. IMO the previous way of passing it in the runtime was better.

runtime/avn/src/lib.rs
```diff
@@ -679,6 +683,7 @@ parameter_types! {
     pub const StringLimit: u32 = 50;
     pub const MetadataDepositBase: Balance = 1 * MILLI_AVT;
     pub const MetadataDepositPerByte: Balance = 100 * MICRO_AVT;
+    pub const DefaultCheckpointFee: Balance = 100 * MILLI_AVT;
```
nahuseyoum, line 686:
```suggestion
    pub const DefaultCheckpointFee: Balance = 60 * MILLI_AVT;
```

nahuseyoum (approved):

pallets/avn-anchor/src/lib.rs
```diff
@@ -377,7 +446,7 @@ pub mod pallet {
             ));
 
             <Nonces<T>>::mutate(chain_id, |n| *n += 1);
-
+            let _ = Self::charge_fee(handler.clone(), chain_id)?;
```
thadouk, line 449:
```suggestion
            Self::charge_fee(handler.clone(), chain_id)?;
```

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +319,47 @@ pub mod pallet {
...
+
+            CheckpointFee::<T>::insert(chain_id, new_fee);
+            Self::deposit_event(Event::CheckpointFeeUpdated { chain_id, new_fee });
+
+            Ok(())
+        }
     }
 
     impl<T: Config> Pallet<T> {
+        pub(crate) fn charge_fee(handler: T::AccountId, chain_id: ChainId) -> DispatchResult {
```
thadouk, line 340: when a different token than the native is used to charge a fee for a checkpoint of a chain, does that mean that we expect to pay 0 amount of that token or standard fees plus the checkpoint fee as a tip?
What happens in the case of `submit_checkpoint_with_identity`?

pallets/avn-anchor/src/lib.rs
```diff
@@ -288,9 +319,47 @@ pub mod pallet {
...
+
+            Ok(())
+        }
+
+        pub fn get_checkpoint_fee(chain_id: ChainId) -> BalanceOf<T> {
+            let fee = CheckpointFee::<T>::get(chain_id);
+            if CheckpointFee::<T>::contains_key(chain_id) {
+                fee
+            } else {
+                T::DefaultCheckpointFee::get()
```
thadouk, line 359: I wonder if we could make the CheckpointFee a ValueQuery with T::DefaultCheckpointFee::get() being the default value, to simplify this and eliminate the helper function.

pallets/avn-anchor/src/tests.rs
```diff
@@ -614,13 +600,164 @@ fn checkpoint_id_overflow_fails() {
...
+
+        assert_ok!(AvnAnchor::submit_checkpoint_with_identity(
+            RuntimeOrigin::signed(handler),
+            checkpoint
+        ));
+
+        System::assert_has_event(
+            Event::CheckpointSubmitted(handler, chain_id, 0, checkpoint).into(),
+        );
+        System::assert_has_event(Event::CheckpointFeeCharged { handler, chain_id, fee }.into());
```
thadouk, line 677: should we assert the sender balances here?

pallets/avn-anchor/src/tests.rs
```diff
@@ -614,13 +600,164 @@ fn checkpoint_id_overflow_fails() {
...
```
thadouk, line 701: shouldn't we assert that the balance of the user was unchanged?
follow up:
zero fees meaning, not even gas fees or 0 check-pointing fee?

thadouk (approved):
do we have fee tests for `signed_submit_checkpoint_with_identity` entrypoint?

pallets/avn-anchor/src/lib.rs
ivan-cholakov: We pay fees in the native token only with the current implementation

pallets/avn-anchor/src/lib.rs
ivan-cholakov: will try it out, thanks

pallets/avn-anchor/src/lib.rs
thadouk: So lets assume that the checkpoint fee is set to 1 NativeToken (NT) and that every tx standard_fee os 0.01 NT
```
total_fees = 0.01 NT + 1 NT = 1.01 NT
```
Is that correct?

ivan-cholakov: yes

PR #450, opened 2024-11-05
feat: upgrade references to polkadot 1.6

thadouk:
## Proposed changes

Upgrade the project references to use version 1.6 of the polkadot-sdk.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2023-05-22"
+channel = "1.74"
 components = [
 	"cargo",
 	"clippy",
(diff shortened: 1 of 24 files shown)


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2023-05-22"
+channel = "1.74"
```
ivan-cholakov, line 2: we need to update also the workflow files to have the same version because of the linting check

ivan-cholakov (approved):

rust-toolchain.toml
thadouk: i think fmt needs nightly to have all the options we use.

PR #464, opened 2024-12-23
Enable ECDSA verification against derived Sr25519 public keys

fluorostani:
## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


primitives/avn-common/src/lib.rs
```diff
@@ -239,19 +239,41 @@ pub fn hash_with_ethereum_prefix(hex_message: &String) -> Result<[u8; 32], ECDSA
...
+            let hashed_ecdsa_pub_key = keccak_256(ecdsa_pub_key.as_ref());
+            let derived_account = sr25519::Public::from_raw(hashed_ecdsa_pub_key);
+
+            if derived_account.encode() == proof.signer.clone().into().encode() {
+                return Ok(())
+            }
+        }
+    }
+
+    Err(())
```
thadouk, line 276: not sure if we want this behaviour to be enabled by default on all chains.
Wrapping with a feature flag this behaviour and enabling where needed is cleaner.

thadouk (approved):

primitives/avn-common/src/lib.rs
```diff
@@ -239,19 +239,41 @@ pub fn hash_with_ethereum_prefix(hex_message: &String) -> Result<[u8; 32], ECDSA
...
+
+    let encoded_signature = proof.signature.encode();
+    if let Ok(ecdsa_signature) = ecdsa::Signature::try_from(&encoded_signature[..]) {
+        let signed_payload_str = String::from_utf8_lossy(signed_payload).to_string();
+
+        if let Ok(ecdsa_pub_key) =
+            recover_public_key_from_ecdsa_signature(&ecdsa_signature, &signed_payload_str)
+        {
+            let hashed_ecdsa_pub_key = keccak_256(ecdsa_pub_key.as_ref());
+            let derived_account = sr25519::Public::from_raw(hashed_ecdsa_pub_key);
```
thadouk, line 268:
```
	/// A new instance from the given 32-byte `data`.
	///
	/// NOTE: No checking goes on to ensure this is a real public key. Only use it if
	/// you are certain that the array actually is a pubkey. GIGO!
	pub fn from_raw(data: [u8; 32]) -> Self {
		Public(data)
	}
```
The hash of the public key is not guaranteed to be a public key. Maybe instead we could try to use the actual public key (32 bytes version) that we have recovered

nahuseyoum (approved):

fluorostani (commented):
All good, thanks @nahuseyoum

PR #468, opened 2025-01-08
update ethereum hasher

nahuseyoum:
## Proposed changes

This PR updates the hashing algorith to match external ethereum wallets.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/pallets/avn-proxy/src/tests/mock.rs
+++ b/pallets/avn-proxy/src/tests/mock.rs
@@ -14,8 +14,7 @@ use pallet_avn::BridgeInterfaceNotification;
 use pallet_balances;
 use pallet_nft_manager::nft_data::Royalty;
 use pallet_session as session;
-use sp_avn_common::hash_with_ethereum_prefix;
-use sp_core::{blake2_256, ecdsa, keccak_256, sr25519, ConstU32, ConstU64, Pair, H160, H256};
+use sp_core::{blake2_256, ecdsa, sr25519, ConstU32, ConstU64, Pair, H160, H256};
 use sp_keystore::{testing::MemoryKeystore, KeystoreExt};
 use sp_runtime::{
     testing::{TestXt, UintAuthorityId},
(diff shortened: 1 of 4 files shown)


pallets/avn-proxy/src/tests/mock.rs
```diff
@@ -389,18 +388,19 @@ impl TestAccount {
         return ecdsa::Pair::from_seed(&self.seed)
     }
 
-    pub fn ethereum_ecdsa_sign(&self, message_to_sign: &[u8]) -> Signature {
-        let hashed_message = hash_with_ethereum_prefix(&hex::encode(message_to_sign)).unwrap();
-        return Signature::from(self.ecdsa_key_pair().sign(&hashed_message))
+    pub fn ethereum_ecdsa_sign(&self, _message_to_sign: &[u8]) -> Signature {
+        // TODO: Use a proper library to sign
```
thadouk, line 392: why not ethers.rs?

primitives/avn-common/src/lib.rs
```diff
@@ -264,6 +278,38 @@ pub fn hash_with_ethereum_prefix(hex_message: &String) -> Result<[u8; 32], ECDSA
     Ok(hash)
 }
 
+// Be careful when changing this logic because it needs to be compatible with other Ethereum
+// wallets. The message_bytes are converted to hex first then to utf8bytes.
+pub fn hash_string_data_with_ethereum_prefix(
```
thadouk, line 283: Would it be possible to use this across the AvN?

thadouk (commented):
Can we get a better description please that highlights why the changes are needed and why the previous signing approach is not enough?

primitives/avn-common/src/lib.rs
nahuseyoum: No, we should use the existing one. This one should be used if we need to validate an externally signed signature.

pallets/avn-proxy/src/tests/mock.rs
nahuseyoum: Yeah, or any other that is easier to integrate.

PR #469, opened 2025-01-08
feat: Add support for ERC20 direct token transfers and primary events

thadouk:
## Proposed changes

- Extend data structures to handle ERC20 `Transfer` events.
- Introduce primary events that override secondary events within the same transaction.
- Enhance `ValidEvents` enum to be iterable for easier maintenance.
- Set tier-2 receiver address as the bridge contract address padded to 32 bytes.
- Limit secondary event support to the EthereumBridge pallet only.
- New event will only be supported by EthBridge pallet.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release


primitives/avn-common/src/event_types.rs
```diff
@@ -308,67 +350,66 @@ impl LiftedData {
...
                 .try_into()
                 .expect("Slice is the correct size"),
         );
-        return Ok(LiftedToPredictionMarketData { token_contract, receiver_address, amount })
+        return Ok(LiftedData {
+            token_contract: H160::zero(),
+            sender_address,
+            // This must be overwritten after the lift reach consensus with the account configured
+            // on the pallet
+            receiver_address: bridge_contract.into(),
```
nahuseyoum, line 409: This makes me a bit nervous because we risk locking the funds in an account without any private key. Can we "default" it to the treasury maybe?

primitives/avn-common/src/event_types.rs
```diff
@@ -698,13 +739,15 @@ pub enum EventData {
...
-    LogLiftedToPredictionMarket(LiftedToPredictionMarketData),
+    LogLiftedToPredictionMarket(LiftedData),
+    LogErc20Transfer(LiftedData),
 }
 
 impl EventData {
     #[allow(unreachable_patterns)]
     pub fn is_valid(&self) -> bool {
         return match self {
+            // LogLowerClaimed missing. TODO add and remove unreachable patterns.
```
nahuseyoum, line 750: I am assuming we never noticed because lower uses the new flow (has never worked with EthereumEvents)?

nahuseyoum (approved):

primitives/avn-common/src/event_types.rs
thadouk: We could, but it's cleaner if this is done in runtime, by the processor that handles the event.
To have the treasury at the client level we will need to extend the runtime API which has a lot of overhead

primitives/avn-common/src/event_types.rs
thadouk: yes, exactly.

micaelffrancoAV (approved):

fluorostani (approved):

PR #477, opened 2025-01-28
feat: Introduces automatic lifts of ERC-20 token transfers to bridge contract

thadouk:
## Proposed changes

- Extends client and runtime to support discovery of direct ERC-20 transfers to the bridge contract.
- Identifies primary and secondary events, ensuring only transfers not associated with a primary event are processed.
- Event filtering is a responsibility of the collators, which submit validated data to the chain.
- Adds configurable vault account to hold the directly lifted tokens

Related Jira ticket:
- SYS-4349

## Type of change/Merge

- [x] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -767,7 +767,7 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "avn-lower-rpc"
-version = "6.7.0"
+version = "6.7.1"
 dependencies = [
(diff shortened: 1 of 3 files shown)


nahuseyoum (approved):
I am assuming this is identical to the other client changes we have done so far.

fluorostani (approved):

PR #478, opened 2025-01-29
SYS-4384: code comments for authors pallet

ivan-cholakov:


pallets/authors-manager/src/lib.rs
```diff
@@ -453,17 +456,7 @@ impl<T: Config> Pallet<T> {
...
-            (b"bytes32".to_vec(), author_id_bytes.to_vec()),
-            (b"bytes".to_vec(), decompressed_eth_public_key.to_fixed_bytes().to_vec()),
-        ];
-        let tx_id = <T as pallet::Config>::BridgeInterface::publish(
-            function_name,
-            &params,
-            PALLET_ID.to_vec(),
-        )
-        .map_err(|e| DispatchError::Other(e.into()))?;
+        let tx_id = Self::publish_to_bridge(&decompressed_eth_public_key, &author_id_bytes)?;
```
nahuseyoum, line 459: publish_to_bridge function always assumes `function_name = b"addAuthor";` We should take this as an argument.

pallets/authors-manager/src/lib.rs
```diff
@@ -520,6 +513,48 @@ impl<T: Config> Pallet<T> {
...
+                    });
+                },
+                AuthorsActionType::Resignation | AuthorsActionType::Slashed => {
+                    Self::remove_ethereum_public_key_if_required(&account_id);
+                    Self::deposit_event(Event::<T>::AuthorDeregistered {
+                        author_id: account_id.clone(),
+                    });
+                },
+                _ => {},
+            }
```
nahuseyoum, line 549: Isn't this done before sending the transaction to Ethereum?

nahuseyoum (changes requested):

pallets/authors-manager/src/lib.rs
```diff
@@ -365,18 +378,9 @@ impl<T: Config> Pallet<T> {
...
-            (b"bytes32".to_vec(), author_id_bytes.to_vec()),
-        ];
-        let tx_id = <T as pallet::Config>::BridgeInterface::publish(
-            function_name,
-            &params,
-            PALLET_ID.to_vec(),
-        )
-        .map_err(|e| DispatchError::Other(e.into()))?;
+        let tx_id =
+            Self::publish_to_bridge(&decompressed_eth_public_key, &author_id_bytes, b"add_author")?;
```
nahuseyoum, line 383: Unless I am missing something, the order of the params between `addAuthor` and `removeAuthor` are not the same which would break the encoding and invalidate signatures on Tier1. If we want to centralise the function, maybe we should pass in the `params` vector instead of the individual fields.

And perhaps we should have tests to ensure we don't accidentally break this logic as its the second time we are finding issues.

pallets/authors-manager/src/lib.rs
```diff
@@ -390,6 +394,34 @@ impl<T: Config> Pallet<T> {
...
+    // ) -> Result<u32, DispatchError> {
+    //     let function_name = b"addAuthor";
+    //     let params = vec![
+    //         (b"bytes".to_vec(), decompressed_key.to_fixed_bytes().to_vec()),
+    //         (b"bytes32".to_vec(), author_id_bytes.to_vec()),
+    //     ];
+
+    //     <T as pallet::Config>::BridgeInterface::publish(function_name, &params,
+    // PALLET_ID.to_vec())         .map_err(|e| DispatchError::Other(e.into()))
+    // }
```
nahuseyoum, line 423:
```suggestion
```

nahuseyoum (changes requested):

nahuseyoum (approved):

fluorostani (approved):

pallets/authors-manager/src/lib.rs
```diff
@@ -390,6 +407,35 @@ impl<T: Config> Pallet<T> {
...
+
+        let params = match operation {
+            AuthorOperationTier1Endpoint::Remove => vec![
+                (b"bytes32".to_vec(), author_id_bytes.to_vec()),
+                (b"bytes".to_vec(), decompressed_eth_public_key.to_fixed_bytes().to_vec()),
+            ],
+            AuthorOperationTier1Endpoint::Add => vec![
+                (b"bytes".to_vec(), decompressed_eth_public_key.to_fixed_bytes().to_vec()),
+                (b"bytes32".to_vec(), author_id_bytes.to_vec()),
+            ],
```
thadouk, line 428: this can be improve, by having a function in the enumeration, say `convert_to_params_data` ,that takes these two arguments and returns the bytes.
so the code would become, `operation.convert_to_params_data(author_id_bytes, decompressed_eth_public_key)`

thadouk (approved):

PR #479, opened 2025-02-03
SYS-4356 Update key derivation and fix ecdsa test signing

fluorostani:
## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/pallets/avn-proxy/Cargo.toml
+++ b/pallets/avn-proxy/Cargo.toml
@@ -31,6 +31,7 @@ frame-system = { default-features = false, git = "https://github.com/paritytech/
 frame-benchmarking = { default-features = false, git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0", optional = true }
 
 [dev-dependencies]
+libsecp256k1 = { version = "0.7.0", default-features = false, features =["hmac","static-context"]}
 hex = "0.4"
 sp-io = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0" }
 sp-keystore = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0" }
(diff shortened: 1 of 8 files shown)


nahuseyoum (approved):

pallets/avn-proxy/Cargo.toml
```diff
@@ -60,6 +61,8 @@ std = [
 	"sp-avn-common/std",
 	"frame-support/std",
 	"frame-system/std",
+	"libsecp256k1/std",
+
```
thadouk, line 65:
```suggestion
```

pallets/avn-proxy/src/tests/proxy_tests_no_fees.rs
```diff
@@ -86,7 +86,7 @@ mod proxy_without_fees {
 
                 assert_eq!(result, expected_eth_address[..]);
 
-                let public_key_bytes: [u8; 32] = hex!("6fc436cd5273e4ecfd09079c982e2824825d5903cb2cf6cfd02586fbf6faa9a3");
+                let public_key_bytes: [u8; 32] = hex!("844da93c328fc981b1c507a926bb64055f01ec6a7b4d1ab9dc79e086655d368d");
```
thadouk, line 89: Can we use a const to store this?
Can we avoid using a raw public key instead of deriving it?

thadouk (approved):

pallets/avn-proxy/src/tests/proxy_tests_no_fees.rs
fluorostani: In this test we're confirming the derivation in the function is correct, so we do need to hard-code the expected value. 

In the other test (`call_is_proxied_with_good_parameters_ecdsa`) however we rely on the correct key being derived by the function, so for that we accept whatever account gets generated. 

Between the two though we have both avenues covered.

PR #486, opened 2025-03-14
feat: extend the client to include additional events on discovery

thadouk:
## Proposed changes

Extends the client to take into consideration the new api of additional events available from the runtime.
Upon receiving a list, the client will perform a search for that event and include it in the results if:
- the event is found
- its signature is part of the allowed ones
- the block number of the event is lower than the active range start limit

Jira ticket:
- SYS-4453

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -1,5 +1,5 @@
 use crate::{web3_utils, BlockT, ETH_FINALITY};
-use futures::{future::join_all, lock::Mutex};
+use futures::{future::try_join_all, lock::Mutex};
 use node_primitives::AccountId;
 use pallet_eth_bridge_runtime_api::EthEventHandlerApi;
 use sc_client_api::{BlockBackend, UsageProvider};
@@ -359,12 +359,12 @@ pub async fn identify_additional_events(
             }
         });
 
-    let results: Vec<Result<Option<DiscoveredEvent>, AppError>> = join_all(futures).await;
+    let results = try_join_all(futures).await?;
 
     // check results, return early if any error occurred
     let mut additional_events = Vec::new();
     for result in results {
-        match result? {
+        match result {
             Some(event) => additional_events.push(event),
             None => {},
         }
@@ -786,8 +786,7 @@ where
         .additional_events(config.client.info().best_hash)
         .map_err(|err| format!("Failed to query event signatures: {:?}", err))?
         .iter()
-        .map(|events_set| events_set.iter().collect::<Vec<_>>())
-        .flatten()
+        .flat_map(|events_set| events_set.iter())
         .cloned()
         .collect();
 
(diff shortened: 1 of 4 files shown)


node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -315,6 +315,55 @@ pub async fn identify_events(
...
+) -> Result<Vec<DiscoveredEvent>, AppError> {
+    // Create a future for each event
+    let futures = additional_events_to_check
+        .iter()
+        .filter(|event| {
+            event.block < start_block &&
+                event_signatures_to_find.contains(&event.event_id.signature)
+        })
+        .map(|event| {
+            let contract = contract_addresses.clone();
```
nahuseyoum, line 334: do we need to clone here?

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -315,6 +315,55 @@ pub async fn identify_events(
...
+                    new_event.event.event_id.transaction_hash == event.event_id.transaction_hash
+                }))
+            }
+        });
+
+    let results: Vec<Result<Option<DiscoveredEvent>, AppError>> = join_all(futures).await;
+
+    // check results, return early if any error occurred
+    let mut additional_events = Vec::new();
+    for result in results {
```
nahuseyoum, line 357: if we want to return early, wouldn't it be easier to use `try_join_all`?

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -721,6 +770,17 @@ where
         )
         .map_err(|err| format!("Failed to check if author has casted event vote: {:?}", err))?;
 
+    let additional_events: Vec<AdditionalEvent> = config
+        .client
+        .runtime_api()
+        .additional_events(config.client.info().best_hash)
+        .map_err(|err| format!("Failed to query event signatures: {:?}", err))?
+        .iter()
+        .map(|events_set| events_set.iter().collect::<Vec<_>>())
```
nahuseyoum, line 779: `fn additional_events()` returns an `Option<AdditionalEvents>`. AdditionalEvents is a `BoundedBTreeSet<AdditionalEvent, ConstU32<16>>` and `AdditionalEvent` is a struct. So what is an `events_set`?

nahuseyoum: If we have an inner array, maybe using `.flat_map(|events_set| events_set.iter())` instead of `map` and `flatten` would make it more efficient and easier to understand.

nahuseyoum (approved):
Not related to this PR but `EventRegistry::new()` function has a duplicate entry for `ValidEvents::LiftedToPredictionMarket.signature()` we should remove one of them.

node/avn-service/src/ethereum_events_handler.rs
thadouk: yes cause we have any async closure that we pass that in.

node/avn-service/src/ethereum_events_handler.rs
thadouk: you can iterate an option.
event_set is an `AdditionalEvents` object which is a BoundedBTreeSet.

PR #487, opened 2025-03-16
feat: Adds support for additional events

thadouk:
## Proposed changes
Extends the client implementation to support additional events during a range processing. Additional events must be from previous blocks of the range.
Extends the runtime API to provide these and the client implementation to consume it.
Adding new events is a sudo operation which is performed with a  new sudo extrinsic that consolidates previous entry points.

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [x] Minor release 
    - [ ] Patch release

### DIFF
--- a/pallets/eth-bridge/src/lib.rs
+++ b/pallets/eth-bridge/src/lib.rs
@@ -1306,7 +1306,7 @@ impl<T: Config> Pallet<T> {
     fn ethereum_event_has_already_been_accepted(tx_hash: &H256) -> bool {
         if let Some(processed_event) = ProcessedEthereumEvents::<T>::get(tx_hash) {
             if processed_event.accepted {
-                return true;
+                return true
             }
         }
         false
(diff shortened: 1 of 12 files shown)


pallets/eth-bridge/src/lib.rs
```diff
@@ -715,6 +715,44 @@ pub mod pallet {
 
             Ok(Some(final_weight).into())
         }
+
+        #[pallet::call_index(8)]
+        #[pallet::weight(Weight::zero())]
```
nahuseyoum, line 720: am guessing the other PR fixes this?

nahuseyoum (approved):

pallets/eth-bridge/src/lib.rs
thadouk: yes

micaelffrancoAV (approved):

PR #498, opened 2025-04-11
DEVOPS-1089 Add build step for AvN node on ubuntu 2404

vukomir:
## Proposed changes
- added build steps for ubuntu 2404
- removed artefacts cleanup and added artifact expiration to 1day
- updated workflows to support gh runner config
- updated Dockerfile to use ubuntu 2404

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -99,7 +99,6 @@ jobs:
       (needs.test-code.result == 'success' || needs.test-code.result == 'skipped')
     uses: ./.github/workflows/docker-build.yml
     with:
-      runs_on: gh-ubuntu-2404-x64
       parachain-collator-artifact: parachain-collator
       parachain-runtime-artifact: parachain-runtime
       release: ${{ needs.set-up-job.outputs.RELEASE }}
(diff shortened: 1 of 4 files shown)


nahuseyoum (approved):

thadouk (commented):
I think that all pipelines that test compilation, unit test and benchmarks are safe to be migrated to 24.04 runners rather than duplicate them.
In fact I think it would make sense to have everything running by default to 24.04 and only replicate the binary build to 20.04, which is also what we are going to use during this transistion period.

SZYanev (approved):

SZYanev (approved):

thadouk (approved):

.github/workflows/cargo-build.yml
```diff
@@ -53,22 +73,25 @@ jobs:
         if: ${{ inputs.upload_artifacts }}
         uses: actions/upload-artifact@v4
         with:
-          name: avn-parachain-collator
+          name: ${{ inputs.parachain-collator-artifact }}
           path: target/release/avn-parachain-collator
           if-no-files-found: error
+          retention-days: 1
```
thadouk, line 79: if possible lets increase these to a few more days

PR #501, opened 2025-04-22
fix: update default AuthorityId in AVN pallet

thadouk:
## Proposed changes

The UintAuthorityId mock is not suitable as the default for the test implementation because it's not available in non-test configurations, which the default implementation relies on.

This commit sets the default AuthorityId to sr25519. Mocks explicitly restore UintAuthorityId where needed.


micaelffrancoAV (approved):

nahuseyoum (commented):
:+1:

PR #504, opened 2025-04-28
feat: Updates client to use polkadot-sdk 1.6 libraries

thadouk:
## Proposed changes

Updates the client to be compliant with polkadot-1.6 libraries.

The client still uses the deprecated way of generating chain specs. This is due to the new way using a JSON object for genesis state creation, uses serde-json to serialise the chainspec object with `arbitrary_precision` feature for u128 integer representations and ERC20 denominations.

This PR is mutually exclusive of #453


thadouk:
Went with the solution on #453 provisionally

PR #507, opened 2025-05-08
fix: ensure stable ordering for Ethereum event partitions

thadouk:
## Proposed changes

Fixes an issue in the Ord implementation of DiscoveredEvent where event
ordering was based on the event signature instead of the transaction hash.
This caused inconsistent partitioning of the same dataset due to unstable
or incorrect ordering.

Ordering is now based on block number and transaction hash to guarantee
uniqueness and determinism.

Extends the set_admin_setting extrinsic to provide the option to reset
the vote data for ethereum events range.

Jira ticket:
- SYS-4548

## Type of change/Merge

- [x] Release 
  - [x] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release

### DIFF
--- a/pallets/eth-bridge/src/lib.rs
+++ b/pallets/eth-bridge/src/lib.rs
@@ -750,7 +750,7 @@ pub mod pallet {
                     Self::deposit_event(Event::<T>::AdditionalEventQueued { transaction_hash });
                 },
                 AdminSettings::RestartEventDiscoveryOnRange => {
-                    let _ = EthereumEvents::<T>::clear(10, None);
+                    let _ = EthereumEvents::<T>::clear(100, None);
                 },
             }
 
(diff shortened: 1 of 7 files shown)


pallets/eth-bridge/src/lib.rs
```diff
@@ -749,6 +749,9 @@ pub mod pallet {
                     .map_err(|_| Error::<T>::QuotaReachedForAdditionalEvents)?;
                     Self::deposit_event(Event::<T>::AdditionalEventQueued { transaction_hash });
                 },
+                AdminSettings::RestartEventDiscoveryOnRange => {
+                    let _ = EthereumEvents::<T>::clear(10, None);
```
nahuseyoum, line 753: I am assuming 10 is how many keys we want to clear? If so, can we increase it in case we have more collators in the future?

nahuseyoum (approved):

PR #511, opened 2025-05-12
fix: manual build step workflow

vukomir:
## Proposed changes
fixed manual build step workflow

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/.github/workflows/manual-build.yaml
+++ b/.github/workflows/manual-build.yaml
@@ -70,7 +70,7 @@ jobs:
     secrets: inherit
 
   build-docker:
-    needs: [ set-up-job, build-code-ubuntu2404 ]
+    needs: [ set-up-job, build-code-ubuntu2404, test-code ]
     if: |
       always() &&
       (needs.build-code-ubuntu2404.result == 'success') &&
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-build.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
     secrets: inherit
 
   build-docker:
-    needs: [ set-up-job, build-code-ubuntu2404 ]
+    needs: [ set-up-job, build-code-ubuntu2404, test-code ]
```
thadouk, line 73: do we want this this? Sometimes we want some builds on feature branches that work is not complete and tests might fail

thadouk (approved):

nahuseyoum (approved):

.github/workflows/manual-build.yaml
thadouk: isn't this does this opposite? makes build-docker need to test the code?

PR #518, opened 2025-05-14
fix: Remove manual rust installation step

thadouk:
## Proposed changes

Rust version should be mandated from rust-toolchain file for build operations.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/.github/workflows/cargo-build.yml
+++ b/.github/workflows/cargo-build.yml
@@ -38,16 +38,6 @@ jobs:
       - name: Check out
         uses: actions/checkout@v3
 
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly-2023-05-22
-          components: rustfmt, clippy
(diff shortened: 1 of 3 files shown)


.github/workflows/cargo-build.yml
```diff
@@ -38,16 +38,6 @@ jobs:
       - name: Check out
         uses: actions/checkout@v3
 
-      - name: Install Rust
```
nahuseyoum, line 41: Does this mean all the workflow files will start using the toolchain defined in the root (in the toolchain file)?

nahuseyoum (approved):

micaelffrancoAV (approved):

.github/workflows/cargo-build.yml
thadouk: not all, the ones define another toolchain, with override, that version will be used.

PR #524, opened 2025-05-23
DEVOPS-1121 Add Ubuntu 2404 to avn-parachain release notes

vukomir:
## Proposed changes

Updated release notes with ubuntu 2404 artifacts

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1416,7 +1416,7 @@ dependencies = [
 [[package]]
 name = "binary-merkle-tree"
 version = "4.0.0-dev"
-source = "git+https://github.com/paritytech/polkadot-sdk?branch=release-polkadot-v1.6.0#f00a911c52c047db3119dedfddc86c332ff6626b"
+source = "git+https://github.com/paritytech/polkadot-sdk?branch=release-polkadot-v1.6.0#e36ffb3da50afff65f8ec99378fbfe14f9d340e7"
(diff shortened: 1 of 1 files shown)


nahuseyoum (approved):

.github/workflows/gh-release.yml
```diff
@@ -38,26 +38,44 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: parachain-runtime
+          path: ubuntu-2404/target/release/wbuild/avn-parachain-runtime/
+
+      - name: get AvN Parachain test runtime Artifact (Ubuntu 24.04)
+        uses: actions/download-artifact@v4
+        with:
+          name: parachain-test-runtime
+          path: ubuntu-2404/target/release/wbuild/avn-parachain-test-runtime/
```
thadouk, line 75: The release should provide only one version of the wasm, that of 24.04 and their sha256 
Since it's build in web-assembly the outcome can be used across platforms and nodes.
Multiple builds though mess up with the vesrsioning as the build output is not deterministic.

PR #539, opened 2025-06-20
feat: Aligns confirmations with EIP-712 schema

thadouk:
## Proposed changes

Updates the way the eth-bridge pallet produces the hashes for confirmations and aligns it with EIP-712 schema.
The hashing task is delegated to Alloy's data types, SolStruct, which has the mechanism embedded. We just introduce the appropriate data structure for each of the AvN operations.

Jira ticket:
- SYS-4560

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/primitives/avn-common/src/eth/mod.rs
+++ b/primitives/avn-common/src/eth/mod.rs
@@ -94,7 +94,7 @@ pub enum BridgeContractMethod {
 }
 
 sol! {
-    struct PublishRootConfirmationData {
+    struct PublishRoot {
         bytes32 root;
         uint64 expiry;
         uint32 tx_id;
@@ -102,7 +102,7 @@ sol! {
 }
 
 sol! {
-    struct TriggerGrowthConfirmationData {
+    struct TriggerGrowth {
         uint128 rewards;
         uint128 avg_staked;
         uint32 period;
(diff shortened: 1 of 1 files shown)


primitives/avn-common/src/eth/mod.rs
```diff
@@ -123,3 +159,107 @@ impl TryFrom<&[u8]> for BridgeContractMethod {
...
+pub fn eip712_hash<T: SolStruct>(data: &T, domain: &Eip712Domain) -> H256 {
+    let hash: B256 = data.eip712_signing_hash(domain);
+    H256::from_slice(&hash.0)
+}
+
+fn parse_from_utf8<T: std::str::FromStr>(bytes: &[u8]) -> Result<T, ()> {
+    str::from_utf8(bytes).map_err(|_| ())?.parse::<T>().map_err(|_| ())
+}
+
+pub fn create_function_confirmation_hash(
```
nahuseyoum, line 172: I think this function should take `data` and `domain` as a parameter and return the eip_712 hash. The logic of constructing the data should stay in the relevant pallets. And maybe we can call it `get_t1_tx_hash` ?

nahuseyoum (commented):
IMO we should try to decouple the primitives logic from the pallets as much as we can. I know some of the changes are not related to this PR (ex: BridgeContractMethod moving to primitives) but it makes it impossible for other clients to integrate third party pallets without changing Avn primitive code.

fluorostani (approved):

PR #556, opened 2025-07-18
feat: enable support for multiple instances in the client

thadouk:
## Proposed changes

Enables support in the client for multiple RPC nodes and enables the ethereum events handler service to manage multiple web3 connections to networks and decide which connection is the right one for the instance

SYS-4601
SYS-4597
## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -453,7 +453,7 @@ where
     pub keystore: Arc<LocalKeystore>,
     pub keystore_path: PathBuf,
     pub avn_port: Option<String>,
-    pub eth_node_url: String,
+    pub eth_node_urls: Vec<String>,
     // TODO add support for multiple web3 instances on different chains
     pub web3_data_mutex: Arc<Mutex<Web3Data>>,
     pub client: Arc<ClientT>,
@@ -485,11 +485,14 @@ where
             let web3_init_time = Instant::now();
             log::info!("⛓️  avn-service: web3 initialisation start");
 
-            let web3 = setup_web3_connection(&self.eth_node_url);
+            // TODO fixme
+            let eth_web3_url =
+                self.eth_node_urls.first().cloned().unwrap_or_else(|| "".to_string());
+            let web3 = setup_web3_connection(&eth_web3_url);
             if web3.is_none() {
                 log::error!(
                     "💔 Error creating a web3 connection. URL is not valid {:?}",
-                    &self.eth_node_url
+                    &self.eth_node_urls
                 );
                 return Err(server_error("Error creating a web3 connection".to_string()))
             }
--- a/node/avn-service/src/ethereum_events_handler.rs
+++ b/node/avn-service/src/ethereum_events_handler.rs
@@ -23,7 +23,7 @@ use sp_blockchain::HeaderBackend;
 use sp_core::{sr25519::Public, H256 as SpH256};
 use sp_keystore::Keystore;
 use sp_runtime::SaturatedConversion;
-use std::{collections::HashMap, marker::PhantomData, time::Instant};
+use std::{collections::HashMap, time::Instant};
 pub use std::{path::PathBuf, sync::Arc};
 use tide::Error as TideError;
 use tokio::time::{sleep, Duration};
@@ -36,7 +36,7 @@ use web3::{
 use pallet_eth_bridge::{SUBMIT_ETHEREUM_EVENTS_HASH_CONTEXT, SUBMIT_LATEST_ETH_BLOCK_CONTEXT};
 use pallet_eth_bridge_runtime_api::InstanceId;
 
-use crate::{server_error, setup_web3_connection, Web3Data};
+use crate::{get_chain_id_from_provider, server_error, setup_web3_connection, Web3Data};
 use sc_transaction_pool_api::OffchainTransactionPoolFactory;
 
 pub struct EventInfo {
@@ -454,10 +454,8 @@ where
     pub keystore_path: PathBuf,
     pub avn_port: Option<String>,
     pub eth_node_urls: Vec<String>,
-    // TODO add support for multiple web3 instances on different chains
-    pub web3_data_mutex: Arc<Mutex<Web3Data>>,
+    pub web3_data_mutexes: HashMap<u64, Arc<Mutex<Web3Data>>>,
     pub client: Arc<ClientT>,
-    pub _block: PhantomData<Block>,
     pub offchain_transaction_pool_factory: OffchainTransactionPoolFactory<Block>,
 }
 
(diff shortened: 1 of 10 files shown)


node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -660,16 +699,22 @@ where
...
             .query_active_block_range(config.client.info().best_hash, instance_id)
             .map_err(|err| format!("Failed to query bridge contract: {:?}", err))?;
 
-        let web3_data_mutex = config.web3_data_mutex.lock().await;
+        let web3_data_lock = match initialize_web3_connection_for_instance(config, &instance).await
+        {
+            Ok(web3_data) => web3_data,
+            Err(e) => {
+                log::error!("Failed to initialize web3 connection for instance: {:?}", e);
+                continue
```
nahuseyoum, line 713: do we want to continue here?

node/src/cli.rs
```diff
@@ -62,8 +62,8 @@ pub struct AvnRunCmd {
     pub avn_port: Option<String>,
 
     /// URL for connecting with an ethereum node
-    #[arg(long = "ethereum-node-url", value_name = "ETH URL")]
-    pub eth_node_url: Option<String>,
+    #[arg(long = "ethereum-node-url", value_name = "ETH URL", num_args = 0..=5)]
```
nahuseyoum, line 65: Why do we limit it to 5 here?

pallets/eth-bridge/src/migration.rs
```diff
@@ -40,7 +40,7 @@ impl<T: Config<I>, I: 'static> OnRuntimeUpgrade for EthBridgeMigrations<T, I> {
         );
         let mut consumed_weight = Weight::zero();
 
-        if onchain < 2 {
+        if EthBlockRangeSize::<T, I>::get() == 0 {
```
nahuseyoum, line 43: why are we deviating from using the built-in method (pallet version)?

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -473,32 +471,75 @@ where
...
-                    "💔 Error creating a web3 connection. URL is not valid {:?}",
-                    &self.eth_node_url
+                log::info!(
+                    "⛓️  Successfully connected to node: {} with chain ID: {}",
+                    eth_node_url,
+                    web3_chain_id
                 );
-                return Err(server_error("Error creating a web3 connection".to_string()))
+
+                {
```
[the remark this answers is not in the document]
thadouk: critical section. But copilot made a good suggestion about the assignment of the variable.

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -473,32 +471,75 @@ where
...
+            None => log::debug!(
+                "⛓️  No existing web3 connection found for network: {}. Initialising new...",
+                chain_id
+            ),
+        }
 
-            let web3_init_time = Instant::now();
-            log::info!("⛓️  avn-service: web3 initialisation start");
+        // No web3 connection found for network_id, try the rest of the URLs
+        let web3_connection_locks = Mutex::new(());
```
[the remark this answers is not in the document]
thadouk: That is not the intended purpose of the lock, it is for whoever access it later on via the Arc smart pointer.

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -473,32 +471,75 @@ where
...
+                    eth_node_url,
+                    web3_chain_id
                 );
-                return Err(server_error("Error creating a web3 connection".to_string()))
+
+                {
+                    // Lock the mutex to ensure only one thread can alter the web3 connection
+                    let _ = web3_connection_locks.lock();
+
+                    if self.web3_data_mutexes.get(&web3_chain_id).is_some() {
```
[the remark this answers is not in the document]
thadouk: we try to initialiase all rpc urls and we don't know which chain they belong, we might have multiple from the same one provided and we might end up overwritting it.

node/avn-service/src/ethereum_events_handler.rs
```diff
@@ -473,32 +471,75 @@ where
...
             }
+        }
 
-            log::info!("⏲️  web3 init task completed in: {:?}", web3_init_time.elapsed());
-            web3_data_mutex.web3 = web3;
-            Ok(())
-        } else {
-            Err(server_error("Failed to acquire web3 data mutex.".to_string()))
+        log::info!("⏲️  web3 init task completed in: {:?}", web3_init_time.elapsed());
+        match self.web3_data_mutexes.get(&chain_id) {
```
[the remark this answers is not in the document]
thadouk: I tried this originally but it more complicated.
If you exit early then you will need to enter again on the next network that you haven't initialised. Then some of the connections are active and used. So you will need to drain the successful ones and iterate through the rest.
I also thought instead of continuing to build a vector of connections to the same network and use them as redundancy/fallback options but that was for now out of scope and this approach felt simpler and cleaner.
One intialisation of all connections then if you found the right one return it

node/avn-service/src/ethereum_events_handler.rs
thadouk: yes, we have other instances that are waiting.

node/avn-service/src/lib.rs
```diff
@@ -92,11 +91,15 @@ impl<Block: BlockT, ClientT: BlockBackend<Block> + UsageProvider<Block>> Config<
             let web3_init_time = Instant::now();
             log::info!("⛓️  avn-service: web3 initialisation start");
 
-            let web3 = setup_web3_connection(&self.eth_node_url);
+            // Use the first rpc node for web3 operations
+            let eth_web3_url =
+                self.eth_node_urls.first().cloned().unwrap_or_else(|| "".to_string());
```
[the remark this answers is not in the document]
thadouk: I can adjust this to be closer to how it was initially

node/src/cli.rs
thadouk: what limit you want to have?

pallets/eth-bridge/src/migration.rs
thadouk: cause on new instances that are included in latest version, the migration is not run and the value is not set

pallets/eth-bridge/src/lib.rs
```diff
@@ -297,7 +298,7 @@ pub mod pallet {
 
     // The number of blocks that make up a range
     #[pallet::storage]
-    pub type EthBlockRangeSize<T: Config<I>, I: 'static = ()> = StorageValue<_, u32, ValueQuery>;
+    pub type EthBlockRangeSize<T: Config<I>, I: 'static = ()> = StorageValue<_, u32, ValueQuery, ConstU32<DEFAULT_ETH_RANGE>>;
```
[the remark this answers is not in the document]
thadouk: I don't think this is accurate. ConstU32 implements the Get trait

node/src/cli.rs
nahuseyoum: I would not put a limit here

pallets/eth-bridge/src/migration.rs
nahuseyoum: When new instances are added, we would need to do a forkless upgrade which would change the pallet version so we could use the migration convention i think.

node/src/cli.rs
thadouk: I mean you can have up to 16 instances, so how about I put 32 as a limit?

pallets/eth-bridge/src/migration.rs
thadouk: The migration will not be triggered as the new instance has the latest version of the pallet. I leave this as is for not, lets sync if we need to review this.

PR #565, opened 2025-08-12
fix: test runtime chain spec generation

thadouk:
## Proposed changes

Also updates garde local setup to use the dev/local eth keys to support local setup.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release


thadouk:
> shouldn't we split it into two - one for the fix and one for the garde changes?

Can do if you prefer, the garde changes are that we are using the same ethereum public keys we use in dev/local in garde-local and centralise it.

PR #567, opened 2025-08-27
feat: make quorum configurable

thadouk:
Makes the quorum implementation a modular componenent that can be set via a interface in the EthBridge pallet.
Default implementation is added for the AvN pallet which is set as the default handler.

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/primitives/avn-common/src/lib.rs
+++ b/primitives/avn-common/src/lib.rs
@@ -395,3 +395,22 @@ pub struct EthQueryResponse {
     pub data: Vec<u8>,
     pub num_confirmations: u64,
 }
+
+pub trait QuorumPolicy {
+    /// Percentage required for quorum.
+    const QUORUM_PERCENT: u32;
+
+    /// Percentage for supermajority quorum.
+    const SUPERMAJORITY_PERCENT: u32;
+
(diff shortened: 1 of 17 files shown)


primitives/avn-common/src/lib.rs
```diff
@@ -395,3 +395,22 @@ pub struct EthQueryResponse {
...
+
+pub trait QuorumPolicy {
+    /// Percentage required for quorum.
+    const QUORUM_PERCENT: u32;
+
+    /// Percentage for supermajority quorum.
+    const SUPERMAJORITY_PERCENT: u32;
+
+    fn required_for(total: u32) -> u32 {
+        (total * Self::QUORUM_PERCENT + 99) / 100
```
nahuseyoum, line 407: Is it worth having a check for `Self::QUORUM_PERCENT <= 100);` and `Self::SUPERMAJORITY_PERCENT <= 100` just in case the values are set wrong?

nahuseyoum (approved):

PR #570, opened 2025-09-01
feat(avn-service): require signed proof for sign/send operations

thadouk:
## Proposed changes

This change hardens `avn-service` by extending its interface so that all `sign` and `send` operations must include a proof before execution. The proof is a signature of the request data (using the schema defined by the runtime) and is attached as a header to the POST request.

This ensures that only properly authorized requests can trigger signing or sending actions.

References:
- SYS-4664
- M-06

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [ ] Patch release

### DIFF
--- a/pallets/avn/src/lib.rs
+++ b/pallets/avn/src/lib.rs
@@ -384,21 +384,26 @@ impl<T: Config> Pallet<T> {
     pub fn post_data_to_service(
         url_path: String,
         post_body: Vec<u8>,
+        proof_maybe: Option<<T::AuthorityId as RuntimeAppPublic>::Signature>,
     ) -> Result<Vec<u8>, DispatchError> {
-        let request = http::Request::default().method(http::Method::Post).body(vec![post_body]);
+        let mut request = http::Request::default().method(http::Method::Post).body(vec![post_body]);
+        if let Some(proof) = proof_maybe {
+            request = request.add_header("X-Service-Proof", &format!("{:?}", proof));
+        }
 
         return Ok(Self::invoke_external_service(request, url_path)?)
     }
 
     pub fn request_ecdsa_signature_from_external_service(
         data_to_sign: &str,
+        proof_maybe: Option<<T::AuthorityId as RuntimeAppPublic>::Signature>,
     ) -> Result<ecdsa::Signature, DispatchError> {
         let mut url = String::from("eth/sign_hashed_data/");
         url.push_str(data_to_sign);
 
         log::info!(target: "avn-service", "avn-service sign request (ecdsa) for hex-encoded data {:?}", data_to_sign);
 
-        let ecdsa_signature_utf8 = Self::get_data_from_service(url)?;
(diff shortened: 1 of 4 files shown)


micaelffrancoAV (approved):

node/avn-service/src/keystore_utils.rs
```diff
@@ -99,3 +101,19 @@ fn key_phrase_by_type(
...
     }
 }
+
+pub fn authenticate_token(
+    keystore: &LocalKeystore,
+    message_data: &Vec<u8>,
+    signature: sr25519::Signature,
+) -> bool {
+    return keystore.sr25519_public_keys(KeyTypeId(*b"avnk")).into_iter().any(|public| {
+        log::warn!(
```
nahuseyoum, line 111:
```suggestion
        log::debug!(
```

pallets/avn/src/lib.rs
```diff
@@ -384,21 +385,26 @@ impl<T: Config> Pallet<T> {
...
         post_body: Vec<u8>,
+        proof_maybe: Option<<T::AuthorityId as RuntimeAppPublic>::Signature>,
     ) -> Result<Vec<u8>, DispatchError> {
-        let request = http::Request::default().method(http::Method::Post).body(vec![post_body]);
-
+        let mut request = http::Request::default().method(http::Method::Post).body(vec![post_body]);
+        if let Some(proof) = proof_maybe {
+            log::debug!("X-Auth proof: {:?}", proof);
+            let proof_data = encode_to_http_data(&proof);
+            log::debug!("X-Auth proof-data: {:?}", proof_data);
```
nahuseyoum, line 394:
```suggestion
            log::debug!("X-Auth encoded proof data: {:?}", proof_data);
```

pallets/eth-bridge/src/eth.rs
```diff
@@ -14,10 +14,14 @@ use sp_runtime::DispatchError;
 use sp_std::vec;
 
 pub fn sign_msg_hash<T: Config<I>, I: 'static>(
+    author: &Author<T>,
     msg_hash: &H256,
 ) -> Result<ecdsa::Signature, DispatchError> {
-    let msg_hash_string = hex::encode(msg_hash);
-    let confirmation = AVN::<T>::request_ecdsa_signature_from_external_service(&msg_hash_string)?;
+    let msg_data = msg_hash.as_ref().to_vec();
+    let hex_data = hex::encode(&msg_data).into_bytes();
```
nahuseyoum, line 21: Unless I am mistaken, `msg_data` and `hed_data` are both bytes for the same thing. Wouldn't it be better to sign a hex encoded proof and request signature on the original msg_data? In other words, 
- the proof that we sign is the hex encoded data
- the data we collect signatures for stays as the original bytes

PR #580, opened 2025-10-03
fix: preserve tx_id during transaction replay

thadouk:
## Proposed changes

Ensure that replays of a submit request do not alter its tx_id. The tx_id serves as the unique identifier for the request, and must remain the same so that interacting pallets can reliably match outcomes to their original submissions.

Jira ticket:
- SYS-4678

## Type of change/Merge

- [ ] Release 
  - [ ] Increase versions 
  - [ ] Baseline tests passed  
  - Release type:
    - [ ] Major release 
    - [ ] Minor release 
    - [x] Patch release 

## Checklist

  - What does it do?
- [ ] Documentation updated


todvildes (approved):

nahuseyoum (commented):
If we keep the same tx_id, I think this will cause a signature re-use vulnerability when collators are corroborating. Can we please check if this is an issue?

thadouk:
> If we keep the same tx_id, I think this will cause a signature re-use vulnerability when collators are corroborating. Can we please check if this is an issue?

You don't get corraborations till a tx is submitted. Once submitted, a round of corraboration will occur till consensus is reached. Then the tx_id is either marked as succeeded or not. If not you can retry but you will get a different tx_id on the retry. So no replay attack is possible.

